PR #19, opened 2020-05-12
[notification] Add notification type setting

### LINKED ISSUE #1: [feature] Add notification types
Goals:
- [x] Create a model `NotificationType` to categorize notifications. 
- [x] Add a markdown formatted message template which should be rendered as HTML to users on admin dashboard and mail. 

**Project Idea**
>Add the concept of notification type, it may be a list with some default values which can then be extended by each module. 
Each notification type shall have a message which is generated from a configurable text template, the text shall be markdown formatted, the markdown text shall be expanded into HTML when the notification is sent via email or viewed via the JS widget.

**Discussion**
Project idea says to have some kind of list, in my opinion having a model will be more appropriate for this use case. .

pandafy:
- Added configurable notification types
- Added configurable message template
- Updated README and added tests

Related to #1

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -30,6 +30,8 @@ Available features
 - Email notifications
 - Web notifications
 - Configurable email theme
+- `Definition of notification types <#notification-types>`_
+- `Possibility to register new notification types <#registering--unregistering-notification-types>`_
 - TODO: add more
 
 Install development version
@@ -180,17 +182,154 @@ Additional ``notify`` keyword arguments
 +---------------------+-----------------------------------------------------------------------------+
 |  **Parameter**      |                             **Description**                                 |
 +---------------------+-----------------------------------------------------------------------------+
-| ``email_subject``   | Sets subject of email notification to be sent.                              |
+|  ``email_subject``  | Sets subject of email notification to be sent.                              |
 |                     |                                                                             |
 |                     | Defaults to the truncated description.                                      |
 +---------------------+-----------------------------------------------------------------------------+
-| ``url``             | Adds a URL in the email text, eg:                                           |
+|       ``url``       | Adds a URL in the email text, eg:                                           |
 |                     |                                                                             |
 |                     | ``For more information see <url>.``                                         |
 |                     |                                                                             |
 |                     | Defaults to **None**, meaning the above message would                       |
 |                     | not be added to the email text.                                             |
 +---------------------+-----------------------------------------------------------------------------+
+|       ``type``      | Set values of other parameters based on predefined setting                  |
+|                     | ``OPENWISP_NOTIFICATION_TYPES``                                             |
+|                     |                                                                             |
+|                     | Defaults to **None** meaning you need to provide other arguments.           |
++---------------------+-----------------------------------------------------------------------------+
+
+Notification Types
+------------------
+
+**OpenWISP Notifications** simplifies configuring individual notification by using notification types.
+You can think of notification type as a template for notifications. 
+
+These properties can be configured for each notification type: 
+
++------------------+--------------------------------------------------------------------------------+
+|   **Property**   |                         **Description**                                        |
++------------------+--------------------------------------------------------------------------------+
+|      level       | Sets ``level`` attribute of the notification.                                  |
++------------------+--------------------------------------------------------------------------------+
+|      verb        | Sets ``verb`` attribute of the notification.                                   |
++------------------+--------------------------------------------------------------------------------+
+|      name        | Sets display name of notification type.                                        |
++------------------+--------------------------------------------------------------------------------+
+|     message      | Sets ``message`` attribute of the notification.                                | 
++------------------+--------------------------------------------------------------------------------+
+|  email_subject   | Sets subject of the email notification.                                        |
++------------------+--------------------------------------------------------------------------------+
+| message_template | Path to file having template for message of the notification.                  |
++------------------+--------------------------------------------------------------------------------+
+
+.. note::
+    A notification type configuration should contain atleast one of ``message`` or ``message_template``
+    settings. If both of them are present, ``message`` is given preference over ``message_template``. 
+
+Defining ``message_template``
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+You can either extend default message template or write your own from scratch. An example to extend default
+message template is shown below.
+
+.. code-block:: jinja2
+
+    # In templates/openwisp_notifications/your_message_template.md
+    {% extends 'openwisp_notifications/default_message.md' %}
+    {% block body %}
+        {{ notification.target }} has malfunctioned. 
+    {% endblock body %}
+
+.. note::
+    You can access all attributes of the notification using ``notification`` variables in your message template as shown above.
+
+Registering / Unregistering Notification Types
+----------------------------------------------
+
+**OpenWISP Notifications** provides registering and unregistering notifications through utility functions
+``openwisp_notifications.types.register_notification_type`` and ``openwisp_notifications.types.unregister_notification_type``. Using
+these functions you can register or unregister notification types from anywhere in your code. 
+
+register_notification_type
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+This function is used to register a new notification type from anywhere in your code.
+
+Syntax:
+
+.. code-block:: python
+
+    register_notification_type(type_name, type_config)
+
++---------------+--------------------------------------------------------------+
+|   Parameter   |                     Description                              |
++---------------+--------------------------------------------------------------+
+|   type_name   | A ``str`` defining name of the notification type.            |
++---------------+--------------------------------------------------------------+
+|  type_config  | A ``dict`` defining configuration of the notification type.  |
++---------------+--------------------------------------------------------------+
+
+An example usage has been shown below.
+
+.. code-block:: python
+
+    from openwisp_notifications.types.register_notification_type
+
+    # Define configuration of your notification type
+    custom_type = {
+        'level': 'info',
+        'verb': 'added',
+        'verbose_name': 'device added',
+        'message': '{notification.target} was {notification.verb} at {notification.timestamp}',
+        'email_subject' : '[{site.name}] A device has been added'
+    }
+
+    # Register your custom notification type
+    register_notification_type('custom_type', custom_type)
+
+.. note::
+
+    It will raise ``ImproperlyConfigured`` exception if a notification type is already registered
+    with same name(not to be confused with verbose_name).
+
+.. note::
+
+    You can use ``site`` and ``notification`` variables while defining ``message`` and ``email_subject``
+    configuration of notification type. They refer to objects of ``django.contrib.sites.models.Site`
+    and ``openwisp_notifications.models.Notification`` repectively. This allows you to use any of their
+    attributes in your configuration.
+
+unregister_notification_type
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+This function is used to unregister a notification type from anywhere in your code.
+
+Syntax:
+
+.. code-block:: python
+
+    unregister_notification_type(type_name)
+
++---------------+--------------------------------------------------------------+
+|   Parameter   |                     Description                              |
++---------------+--------------------------------------------------------------+
+|   type_name   | A ``str`` defining name of the notification type.            |
++---------------+--------------------------------------------------------------+
+
+An example usage is shown below.
+
+.. code-block:: python
+
+    from openwisp_notifications.types.uregister_notification_type
+
+    # Unregister previously registered notification type
+    unregister_notification_type('custom type')
+
+.. note::
+
+    It will raise ``ImproperlyConfigured`` exception if the concerned notification type is not
+    registered. 
 
 Contributing
 ------------
--- a/openwisp_notifications/apps.py
+++ b/openwisp_notifications/apps.py
@@ -7,7 +7,7 @@ class OpenwispNotificationsConfig(AppConfig):
     verbose_name = _('Notifications')
 
     def ready(self):
-        from openwisp_notifications.signals import notify  # noqa
+        from openwisp_notifications.signals import notify
         from openwisp_notifications.handlers import notify_handler
 
         notify.connect(
--- a/openwisp_notifications/handlers.py
+++ b/openwisp_notifications/handlers.py
@@ -5,6 +5,7 @@ from django.db.models import Q
 from django.db.models.query import QuerySet
 from django.utils import timezone
 from openwisp_notifications import settings as app_settings
+from openwisp_notifications.types import get_notification_configuration
 from swapper import load_model
 
 User = get_user_model()
@@ -14,7 +15,7 @@ EXTRA_DATA = app_settings.get_config()['USE_JSONFIELD']
 Notification = load_model('openwisp_notifications', 'Notification')
 
 
-def notify_handler(verb, **kwargs):
+def notify_handler(**kwargs):
     """
     Handler function to create Notification instance upon action signal call.
     """
@@ -24,9 +25,13 @@ def notify_handler(verb, **kwargs):
     public = bool(kwargs.pop('public', True))
     description = kwargs.pop('description', None)
     timestamp = kwargs.pop('timestamp', timezone.now())
-    level = kwargs.pop('level', Notification.LEVELS.info)
     recipient = kwargs.pop('recipient', None)
-
+    notification_type = kwargs.pop('type', None)
+    notification_template = get_notification_configuration(notification_type)
+    level = notification_template.get(
+        'level', kwargs.pop('level', Notification.LEVELS.info)
+    )
+    verb = notification_template.get('verb', kwargs.pop('verb', None))
     target_org = getattr(kwargs.get('target', None), 'organization_id', None)
 
     where = Q(is_superuser=True)
@@ -51,12 +56,12 @@ def notify_handler(verb, **kwargs):
             .order_by('date_joined')
             .filter(where)
         )
-    new_notifications = []
 
     optional_objs = [
         (kwargs.pop(opt, None), opt) for opt in ('target', 'action_object')
     ]
 
+    new_notifications = []
     for recipient in recipients:
         newnotify = Notification(
             recipient=recipient,
@@ -67,6 +72,7 @@ def notify_handler(verb, **kwargs):
             description=description,
             timestamp=timestamp,
             level=level,
+            type=notification_type,
         )
 
         # Set optional objects
--- a/openwisp_notifications/models.py
+++ b/openwisp_notifications/models.py
@@ -1,13 +1,20 @@
 from django.conf import settings
 from django.contrib.auth import get_user_model
+from django.contrib.sites.models import Site
 from django.core.cache import cache
 from django.core.exceptions import ObjectDoesNotExist
 from django.core.mail import send_mail
 from django.db import models
 from django.db.models.signals import post_delete, post_save
 from django.dispatch import receiver
+from django.template.loader import render_to_string
+from django.utils.functional import cached_property
 from django.utils.translation import ugettext_lazy as _
 from notifications.base.models import AbstractNotification
+from openwisp_notifications.types import (
+    NOTIFICATION_CHOICES,
+    get_notification_configuration,
+)
 from swapper import swappable_setting
 
 from openwisp_utils.base import TimeStampedEditableModel, UUIDModel
@@ -17,6 +24,7 @@ User = get_user_model()
 
 class Notification(UUIDModel, AbstractNotification):
     COUNT_CACHE_KEY = 'ow2-unread-notifications-{0}'
+    type = models.CharField(max_length=30, null=True, choices=NOTIFICATION_CHOICES)
 
     class Meta(AbstractNotification.Meta):
         abstract = False
@@ -30,6 +38,31 @@ class Notification(UUIDModel, AbstractNotification):
     def invalidate_cache(cls, user):
         """ invalidate cache for user """
         cache.delete(cls.COUNT_CACHE_KEY.format(user.pk))
+
+    @cached_property
+    def message(self):
+        if self.type:
+            config = get_notification_configuration(self.type)
+            if 'message' in config:
+                return config['message'].format(notification=self)
+            else:
+                return render_to_string(
+                    config['message_template'], context=dict(notification=self)
+                ).strip()
+        else:
+            return self.description
+
+    @cached_property
+    def email_subject(self):
+        if self.type:
+            config = get_notification_configuration(self.type)
+            return config['email_subject'].format(
+                site=Site.objects.get_current(), notification=self
+            )
+        elif self.data.get('email_subject', None):
+            return self.data.get('email_subject')
+        else:
+            return self.message
 
 
 class NotificationUser(TimeStampedEditableModel):
@@ -75,9 +108,9 @@ def send_email_notification(sender, instance, created, **kwargs):
     ):
         return
     # send email
-    subject = instance.data.get('email_subject', instance.description)
-    url = instance.data.get('url', '')
-    description = instance.description
+    subject = instance.email_subject
+    url = instance.data.get('url', '') if instance.data else None
+    description = instance.message
     if url:
         description += '\n\nFor more information see {0}.'.format(url)
     send_mail(
--- a/openwisp_notifications/tests/test_notifications.py
+++ b/openwisp_notifications/tests/test_notifications.py
@@ -3,11 +3,20 @@ from datetime import timedelta
 from django.contrib.auth import get_user_model
 from django.contrib.contenttypes.models import ContentType
 from django.core import mail
+from django.core.exceptions import ImproperlyConfigured
+from django.template import TemplateDoesNotExist
 from django.test import TestCase
 from django.utils import timezone
 from django.utils.timesince import timesince
 from openwisp_notifications.handlers import notify_handler
 from openwisp_notifications.signals import notify
+from openwisp_notifications.types import (
+    NOTIFICATION_CHOICES,
+    _unregister_notification_choice,
+    get_notification_configuration,
+    register_notification_type,
+    unregister_notification_type,
+)
 from swapper import load_model
 
 from openwisp_users.models import Group, OrganizationUser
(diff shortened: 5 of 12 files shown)


README.rst
```diff
@@ -167,12 +167,43 @@ Additionally Supported Parameters
...
+|                 | ``OPENWISP_NOTIFICATION_TYPE``                                              |
+|                 |                                                                             |
+|                 | Default to **None** meaning you need to provide verb and level in arguments.|
++-----------------+-----------------------------------------------------------------------------+
+
+Settings
+--------
+
+``OPENWISP_NOTIFICATION_TYPE``
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
nemesisdesign, line 186: maybe i'd use the plural form

openwisp_notifications/handlers.py
```diff
@@ -24,9 +25,16 @@ def notify_handler(verb, **kwargs):
     public = bool(kwargs.pop('public', True))
     description = kwargs.pop('description', None)
     timestamp = kwargs.pop('timestamp', timezone.now())
-    level = kwargs.pop('level', Notification.LEVELS.info)
     recipient = kwargs.pop('recipient', None)
 
+    notification_template = NOTIFICATION_TYPES.get(kwargs.pop('type', None), None)
```
nemesisdesign, line 30: I think we need to add a field to the model as well.
Make sure it's indexed with `db_index=True`

pandafy:
I have used `render_to_string` to read message template (currently written in normal text, but can also be written in markdown when we implement markdown to HTML feature) and supply context variables to generate messages for different notifications. I have used template tags which might increase complexity of code, but in my opinion provides a way to easily extend message template and also resembles what Django does with it's templates. If not required I will remove them.

pandafy:
I think failing lines can be covered with a sample app later. :smile:

openwisp_notifications/models.py
```diff
@@ -17,6 +19,12 @@
 
 class Notification(UUIDModel, AbstractNotification):
     COUNT_CACHE_KEY = 'ow2-unread-notifications-{0}'
+    notification_type = models.CharField(
```
nemesisdesign, line 22: This would result in `Notification.notification_type`, which is redundant, I'd rename it to just `type`, don't worry about the base `type()` function having the same name, it's ok if we do it deliberately and we know what we are doing

openwisp_notifications/models.py
```diff
@@ -75,7 +87,7 @@ def send_email_notification(sender, instance, created, **kwargs):
     ):
         return
     # send email
-    subject = instance.data.get('email_subject', instance.description[0:24])
+    subject = instance.data.get('email_subject', instance.short_description[0:24])
```
nemesisdesign, line 90: I don't understand this. Why don't we dedicate a part of the notification configuration to the subject?

openwisp_notifications/notification_types.py
```diff
@@ -0,0 +1,55 @@
...
+    return notification_types
+
+
+def get_notification_types_choices():
+    notification_types = NOTIFICATION_TYPES
+    choices = []
+    for key in sorted(notification_types.keys()):
+        name = notification_types[key].get('name', key)
+        choices.append((key, name))
+    return choices
```
nemesisdesign, line 34: I'm afraid we have to do something different. I realized this after I worked on the chart configuration of the monitoring system
I think this is executed as soon as this module is loaded, hence, the choices available to `Notification.type` will be only those ones who have been registered before.

If another module is loaded after openwisp-monitoring and registers a notification type, I think it will not produce any effect

I suggest to try this that I described above :point_up:.

The first solution that comes into my mind is that we have to add another data structure, `NOTIFICATION_CHOICES`, which is also manipulated by `register_notification_type` and `unregister_notifcation_type`, that way, the choices should be available also if the registration happens after openwisp-monitoring is loaded.

README.rst
```diff
@@ -167,12 +168,85 @@ Additionally Supported Parameters
...
+``OPENWISP_NOTIFICATION_MESSAGE_TEMPLATE``
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
++--------------+--------------------------------+
+| **type**:    | ``string``                     |
++--------------+--------------------------------+
+| **default**: | ``configurables/message.md``   |
++--------------+--------------------------------+
+
+This setting allows to provide a default markdown formatted message template.
```
nemesisdesign, line 220: I think there was a misunderstanding here.

We need 2 things:

1. a general email template, for styling emails
2. each notification type should have its own message (which should be displayed both in email and web notifications) and a subject (the subject should be used in the email subject and can be used the list of the notifications in the JS widget), which can be markdown strings, I think these can go in the `NOTIFICATION_TYPES`

README.rst
```diff
@@ -167,12 +168,85 @@ Additionally Supported Parameters
...
+If a notification type does not define it's message template explicitly, then this message template will be used.
```
nemesisdesign, line 221: I think a notification type must define its message and subject unless we find a use case that forces us to give up this requirement. At the moment I don't see this use case, if you see one, let me know.

README.rst
```diff
@@ -13,8 +13,8 @@ OpenWISP Notifications Module
 It is used to notify OpenWISP users about meaningful events in their network.
 It not only handles common tasks like selecting appropriate recipients for a notification,
 sending email notifications, etc. but also provide measures to customize those notifications with provision for
-configurable email templates, grouping of notifications to ease management and so on. 
-**openwisp-notifications** takes care of all this, so you can focus on what matters the most. 
+configurable email templates, grouping of notifications to ease management and so on.
+**openwisp-notifications** takes care of all this, so you can focus on what matters the most.
```
nemesisdesign, line 17: Start to create a feature list like we have here https://github.com/openwisp/openwisp-ipam#available-features

openwisp_notifications/tests/test_notifications.py
```diff
@@ -152,3 +160,48 @@ def test_organization_recipient(self):
...
+        # Unregistering a notification type which does not exists
+        with self.assertRaises(ImproperlyConfigured):
+            unregister_notifcation_type('wrong type')
+
+        # Unregistering notification type it's name
+        unregister_notifcation_type('test type')
+        self._create_notification()
+        n = notification_queryset.first()
+        self.assertEqual(n.level, 'info')
+        self.assertEqual(n.verb, 'Test Notification')
```
nemesisdesign, line 207: since this test is pretty long, maybe we can use subtests, it should help: https://www.caktusgroup.com/blog/2017/05/29/subtests-are-best/

We should start using more of this, me incldued. This is a good time to start.

@PabloCastellano @nepython @NoumbissiValere @atb00ker take note of this :point_up:

openwisp_notifications/models.py
pandafy: Thanks, that was my only concern to do it like this. I will change it to `type`.

openwisp_notifications/models.py
pandafy: It was implemented like this in `openwisp_monitoring.notifications`. I will add a configurable setting for subject as well. 
**Questions**
- Do we have to truncate length of subject or just inform user in the documentation as best practices for giving subject settings? 
- Do we have make it accept variables as it is done in notification message ?

openwisp_notifications/notification_types.py
pandafy: I am afraid that I don't understand the point of defining choices properly. Is it there to maintain integrity of database and prevent user for defining any arbitrary notification type?

README.rst
pandafy: The words I used might be not proper to convey this. I wanted to say that this setting will provide a template for notification message, not for email. I will try to emphasize that.

README.rst
pandafy: The content of [configurables/message.md](https://github.com/openwisp/openwisp-notifications/pull/19/files#diff-50457a64c07afe611379ffa9e1a20212) is very generic as of now. It can work for every notification. This serves as a fallback setting if `message` is not defined  for a particular notification type. 

Use Case:
For monitoring module which currently implements only two notification types can use a common template for both. This will help us from repeating ourselves while defining a notification type.

openwisp_notifications/tests/test_notifications.py
pandafy: Subtests looks cool, I will rewrite this implementation. :smile:

tests/settings.py
```diff
@@ -120,6 +120,14 @@
     },
 ]
 
+OPENWISP_NOTIFICATION_TYPES = {
+    'dummy': {
+        'level': 'dummy level',
+        'verb': 'dummy verb',
+        'name': 'Dummy Notification Type',
```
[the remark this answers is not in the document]
pandafy: It will use the fallback one defined by `OPENWISP_NOTIFICATION_MESSAGE_TEMPLATE`.

openwisp_notifications/notification_types.py
pandafy: I was looking for a way it can be implemented and found this in [Django Documentation](https://docs.djangoproject.com/en/3.0/ref/models/fields/#choices). 
> Note that choices can be any sequence object – not necessarily a list or tuple. This lets you construct choices dynamically. But if you find yourself hacking choices to be dynamic, you’re probably better off using a proper database table with a ForeignKey. **choices is meant for static data that doesn’t change much, if ever.**

> Note
> A new migration is created each time the order of choices changes.

openwisp_notifications/models.py
pandafy: `django-notifications` provides a `JSONField` - `data`, email's subject is stored in that field.  Would you like to have a dedicated field for the subject?

openwisp_notifications/notification_types.py
nemesisdesign: because notifications will be generated based on their notification types. At some point we should be able to run code like:

`Notification.generate(type='monitoring.device.up')`
`Notification.generate(type='monitoring.device.down')`
`Notification.generate(type='firmware_upgrder.device.upgrade_started')`
`Notification.generate(type='firmware_upgrder.device.upgrade_completed')`
`Notification.generate(type='firmware_upgrder.device.upgrade_failed')`

In order to do this, we need to be able to have other modules extend the types.

We cannot go for the ForeignKey option: that would create migrations in this module, which makes everything incredibly painful. If a user is developing a custom openwisp derivative and starts introducing new notification types, migrations will be created here, then one day we'll change some model detail here and we'll generate other migrations and make their migrations conflict with ours.

In general the stuff that the user can manipulate should go in the database, but in this case the end-users using the OpenWISP application or its derivatives won't have to alter notification types, only we programmers do, we will add new notification types based on the requirements we get from companies and organizations using openwisp.

tests/settings.py
```diff
@@ -120,6 +120,15 @@
...
+        'subject': '[openwisp-notification] Dummy Email Notification',
```
nemesisdesign, line 128: I would remove `[openwisp-notification] `.

I think we should aim to have subjects like:

- `[{site.name}] {device.name} is not reachable`
- `[{site.name}] {device.name} is reachable`

We should strive to ensure that is possible. The variable site is from [django sites](https://docs.djangoproject.com/en/3.0/ref/contrib/sites/).

README.rst
```diff
@@ -24,6 +24,13 @@ configurable email templates, grouping of notifications to ease management and s
 
 ------------
 
+Available Features
+------------------
+
+* Generation and transmission of notifications
+* Receive notifications on email
```
nemesisdesign, line 31: I think this should be `Send notifications via email`

README.rst
```diff
@@ -24,6 +24,13 @@ configurable email templates, grouping of notifications to ease management and s
...
+* Customizable notifications
```
nemesisdesign, line 32: I'd write `Configurable and extensible notification types` and would link a new section where we explain how to register / unregister notification types.

openwisp_notifications/notification_types.py
```diff
@@ -0,0 +1,78 @@
+from django.core.exceptions import ImproperlyConfigured
```
nemesisdesign, line 1: please shorten this file to `types`, imports will then look like `from openwisp_notifications.types import register_notification_type`.

openwisp_notifications/notification_types.py
```diff
@@ -0,0 +1,78 @@
...
+    return choices
+
+
+def validate_notification_type(notification_type):
+    for key, options in notification_type.items():
+        assert 'level' in options
+        assert 'verb' in options
+
+
+def register_notification_type(notification_type):
```
nemesisdesign, line 48: please document the usage of this function.

Test it in openwisp-monitoring to register a notification type there and ensure the notification type appears in the available choices.

**Addendum**

I've done a quick test:

```python
# ... omitting code ...
def ready(self):
    from openwisp_notifications.notification_types import register_notification_type
    register_notification_type({
        'monitoring.device.up': {
            'level': 'info',
            'verb': 'up',
            'name': 'UP',
            'email_subject': '[UP] test',
        }
    })
    register_notification_type({
        'monitoring.device.down': {
            'level': 'warning',
            'verb': 'down',
            'name': 'DOWN',
            'email_subject': '[DOWN] test',
        }
    })
```

Seems to work:

```
$ ./manage.py shell_plus
# Shell Plus Model Imports
# ... omitting output ...
from openwisp_notifications.models import Notification, NotificationUser
# ... omitting output ...

In [1]: Notification.type.field.choices                                                                 
Out[1]: 
[('default', 'Default Type'),
 ('monitoring.device.up', 'UP'),
 ('monitoring.device.down', 'DOWN')]
```

Cool! :+1: :clap: 

What I would change is the syntax:

```python
register_notification_type('monitoring.device.up', {
    'level': 'info',
    'verb': 'up',
    'name': 'UP',
    'email_subject': '[UP] yo',
})
```

openwisp_notifications/notification_types.py
```diff
@@ -0,0 +1,78 @@
...
+    validate_notification_type(notification_type)
+    NOTIFICATION_TYPES.update(notification_type)
+    register_notification_choice(notification_type)
+
+
+def unregister_notification_type(notification_type):
```
nemesisdesign, line 54: please document the usage of this function.

Test it in openwisp-monitoring to unregister a notification type that you register just before and ensure the notification type does not appear in the available choices (this test makes sense only after you get `register_notification_type` to  work as expected.

Can notification types be unregistered with something like this right?

```python
unregister_notification_type( 'monitoring.device.up')
```

openwisp_notifications/notification_types.py
pandafy: Thanks for clearing out my confusion. :smile: 
I hope `generate` method is used for sake of example, since `notify` signal is already handling this. :grimacing:

openwisp_notifications/notification_types.py
pandafy: A similar test exists in `openwisp-notifications`, we can later check this in a sample app. Adding this test again will create redundancy that too in different repositories. Please have a look on [this test case](https://github.com/openwisp/openwisp-notifications/pull/19/files#diff-f5244de33ac705050d99b0ff4150c6a8R184-R202)

I will document usage of these functions.

> Can notification types be unregistered with something like this right?
Yes, they can be unregistered like that.

openwisp_notifications/models.py
nemesisdesign: > Would you like to have a dedicated field for the subject?

Can we get the subject from the type configuration instead? So we don't avoid writing the same content in the DB zillions of time.
The subject will be always the same, the only part that changes are variables and can be passed in.

openwisp_notifications/notification_types.py
nemesisdesign: yes is an example :+1:

openwisp_notifications/notification_types.py
nemesisdesign: > Please have a look on this test case

Are you sure that link is working as intended? It looks like it's not to me.

openwisp_notifications/models.py
```diff
@@ -31,6 +34,16 @@ def invalidate_cache(cls, user):
         """ invalidate cache for user """
         cache.delete(cls.COUNT_CACHE_KEY.format(user.pk))
 
+    @cached_property
+    def short_description(self):
+        return self.description.partition('\n')[0]
```
nemesisdesign, line 39: I think the description can be loaded from the type configuration, like we do with chart configurations, see: 
https://github.com/openwisp/openwisp-monitoring/blob/master/openwisp_monitoring/monitoring/models.py#L298-L358

This way we can avoid repeating the same identical content in the DB.

openwisp_notifications/models.py
```diff
@@ -31,6 +34,16 @@ def invalidate_cache(cls, user):
...
+
+    @property
+    def email_subject(self):
+        if self.data.get('email_subject'):
+            return self.data.get('email_subject')
+        return self.short_description
```
nemesisdesign, line 45: same here

openwisp_notifications/notification_types.py
nemesisdesign: One more thing, see how django implements `register` and `unregister` for the admin:

https://github.com/django/django/blob/master/django/contrib/admin/sites.py#L88-L147

openwisp_notifications/notification_types.py
pandafy: GitHub doesn't open the file if you have already marked it as viewed. Please see those tests here : 

https://github.com/openwisp/openwisp-notifications/blob/927f4fea62e15d2b85ad1485faaadabac03e3248/openwisp_notifications/tests/test_notifications.py#L184-L283

openwisp_notifications/notification_types.py
pandafy: > What I would change is the syntax:
> 
> ```python
> register_notification_type('monitoring.device.up', {
>     'level': 'info',
>     'verb': 'up',
>     'name': 'UP',
>     'email_subject': '[UP] yo',
> })
> ```

I think an iterable will make this syntax a bit messy. The current syntax of having `dict` of `dict` seems more appropriate. It also resembles with definition of `OPENWISP_NOTIFICATION_TYPES`. No doubt the suggested one looks very cleaner.

openwisp_notifications/models.py
pandafy: I tried to follow Laws of Demeter to maintain loose coupling. I have tried to use `NOTIFICATION_TYPES` only in `handlers.py` This was the only reason to implement it like that.

README.rst
```diff
@@ -28,7 +28,9 @@ Available features
 ------------------
 
 - `Generate notifications <#generating-notifications>`_
-- TODO: add more
+- Generation and transmission of notifications
```
nemesisdesign, line 31: this line is a duplicate of line 30

README.rst
```diff
@@ -179,6 +182,61 @@ Additionally Supported Parameters
...
+|       type      | Set values of other parameters based on predefined setting                  |
+|                 | ``OPENWISP_NOTIFICATION_TYPES``                                             |
+|                 |                                                                             |
+|                 | Defaults to **None** meaning you need to provide other arguments.           |
++-----------------+-----------------------------------------------------------------------------+
+
+Settings
+--------
+
+``OPENSWISP_NOTIFICATION_TYPES``
```
nemesisdesign, line 194: I don't think we should make this a setting, it will be managed by register and unregister, right? So we should explain how to use those functions I think, because just changing the setting will have no effect, users of this module will have to use the new notification types in their code, so this is more a developer feature than just a setting. Don't you think?

README.rst
```diff
@@ -28,7 +28,9 @@ Available features
 ------------------
 
 - `Generate notifications <#generating-notifications>`_
-- TODO: add more
+- Generation and transmission of notifications
+- Send notifications via email
+- Configurable and extensible notification types
```
nemesisdesign, line 33: please link the section which explains more info about this point as in line 30

openwisp_notifications/models.py
```diff
@@ -1,22 +1,30 @@
...
+from openwisp_notifications.types import (
+    NOTIFICATION_CHOICES,
+    get_notification_configuration,
+)
 from swapper import swappable_setting
 
 from openwisp_utils.base import TimeStampedEditableModel, UUIDModel
 
 User = get_user_model()
+current_site = Site.objects.get_current()
```
nemesisdesign, line 22: never perform queries to the database in files which are loaded by django at startup (models.py, apps.py and files imported by these two), because this code is executed anytimes you load django, even for doing `./manage.py makemigrations --dry-run`, this is the reason of the build failure in travis.

openwisp_notifications/types.py
```diff
@@ -0,0 +1,90 @@
...
+
+def deep_merge_dicts(dict1, dict2):
+    result = dict1.copy()
+    for key, value in dict2.items():
+        if isinstance(value, dict):
+            node = result.get(key, {})
+            result[key] = deep_merge_dicts(node, value)
+        else:
+            result[key] = value
+    return result
```
nemesisdesign, line 23: can you please open a PR for this in openwisp-utils? So we can pull this code from there. I will then change also openwisp-monitoring to use that one.

openwisp_notifications/templates/configurables/message.md
```diff
@@ -0,0 +1,5 @@
+{% block head %} {{ level }} : {{ verb }} {% endblock head %}
```
nemesisdesign, line 1: I'd change the path of this file to:

`templates/notifications/default_message.md` or `templates/openwisp_notifications/default_message.md`.

`configurables` is unusual.

pandafy:
I don't think users of this module will need register and unregister functions for `NOTIFICATION_CHOICES`. I think it will only create problems if they meddle with them. 

@nemesisdesign I am thinking to combine their logic in register/unregister of `NOTIFICATION_TYPES`. Or we can go Pythonic way of starting their name with `_` ?

README.rst
```diff
@@ -179,6 +182,138 @@ Additionally Supported Parameters
...
+|   description    | Sets ``description`` attribute of the notification.                            | 
++------------------+--------------------------------------------------------------------------------+
+|  email_subject   | Sets subject of the email notification.                                        |
++------------------+--------------------------------------------------------------------------------+
+| message_template | Path to file having template for description of the notification.              |
++------------------+--------------------------------------------------------------------------------+
+
+.. note::
+    A notification type configuration should contain atleast one of ``description`` or ``message_template``
+    settings. If both of them are present, ``description`` is given preference over ``message_template``.
```
nemesisdesign, line 216: what's the difference between `description` and `message_template`?

README.rst
```diff
@@ -179,6 +182,138 @@ Additionally Supported Parameters
...
+    {% block body %}
+        {{ opts.target }} has malfunctioned. 
+    {% endblock body %}
+
+.. note::
+    You can access all attributes of the notification using ``opts`` variables in your message template as shown above.
+
+Registering/Unregistering Notification Types
+--------------------------------------------
+**OpenWISP Notifications** provides registering and unregistering notifications through utility functions
```
nemesisdesign, line 237: leave a blank line, apply to the othe headings.

README.rst
```diff
@@ -179,6 +182,138 @@ Additionally Supported Parameters
...
+
+    from openwisp_notifications.types.register_notification_type
+
+    # Define configuration of your notification type
+    custom_type = {
+            'level': 'info',
+            'verb': 'added',
+            'verbose_name': 'device added',
+            'description': '{opts.target} was {opts.verb} at {opts.timestamp}',
+            'email_subject' : '[{site.name}] A device has been added'
```
nemesisdesign, line 270: indentation looks more than 4 spaces, please ensure it's 4 spaces

README.rst
```diff
@@ -179,6 +182,138 @@ Additionally Supported Parameters
...
+    register_notification_type('custom_type', custom_type)
+
+.. note::
+
+    It will raise ``ImproperlyConfigured`` exception if a notification type is already registered
+    with same name(not to be confused with verbose_name).
+
+.. note::
+
+    You can use ``site`` and ``opts`` variables while defining ``description`` and ``email_subject``
```
nemesisdesign, line 283: `opts` sounds a bit misleading, because it's usually referred for model options, I think `instance`, `obj` or `notification` would be more suited.

openwisp_notifications/models.py
```diff
@@ -31,6 +39,31 @@ def invalidate_cache(cls, user):
         """ invalidate cache for user """
         cache.delete(cls.COUNT_CACHE_KEY.format(user.pk))
 
+    @cached_property
+    def notification_description(self):
```
nemesisdesign, line 43: whati f we call it `message`?

README.rst
pandafy: `description` will be a string with placeholders while `message_template` will be path to template file which will be rendered.

README.rst
```diff
@@ -179,6 +182,142 @@ Additionally Supported Parameters
...
+    # In templates/openwisp_notifications/your_message_template.md
+    {% extends 'openwisp_notifications/default_message.md' %}
+    {% block body %}
+        {{ notification.target }} has malfunctioned. 
+    {% endblock body %}
+
+.. note::
+    You can access all attributes of the notification using ``notification`` variables in your message template as shown above.
+
+Registering/Unregistering Notification Types
```
nemesisdesign, line 236: add a space before and after `/`

README.rst
```diff
@@ -28,7 +28,9 @@ Available features
 ------------------
 
 - `Generate notifications <#generating-notifications>`_
-- TODO: add more
+- Send notifications via email
+- `Configurable and extensible notification types <#registeringunregistering-notification-types>`_
+- `Generalize Notifications <#notification-types>`_
```
nemesisdesign, line 33: one more thing, I don't like these last two lines, I would write something like:

```
- Definition of notification types (link to section)
- Possibility to register new notification types (link to section)
```

openwisp_notifications/apps.py
```diff
@@ -7,9 +8,19 @@ class OpenwispNotificationsConfig(AppConfig):
...
+        self.add_default_menu_items()
+
+    def add_default_menu_items(self):
+        menu_setting = 'OPENWISP_DEFAULT_ADMIN_MENU_ITEMS'
+        items = [{'model': 'openwisp_notifications.Notification'}]
+        if not hasattr(settings, menu_setting):
+            setattr(settings, menu_setting, items)
+        else:
+            current_menu = getattr(settings, menu_setting)
+            current_menu += items
```
nemesisdesign, line 26: I did not spot this, please remove it, we don't need it. We have the bell icon and we'll have the notification JS widget for this.

PR #29, opened 2020-05-29
[models] Added swappable models

### LINKED ISSUE #10: [models] Swappable models
- [ ] Implement Swappable Models
- [ ] Create a sample app to test swappable models
- [ ] Provide documentation about how to extend and customize this module

Useful resources:

- https://github.com/openwisp/openwisp-firmware-upgrader/pull/60 (you can replicate what has been done here)

_Edited by @pandafy_

pandafy:
Removed base admin
Moved all handlers from models.py > handlers.py
Added sample app

Closes #10


openwisp_notifications/tests/test_admin.py
```diff
@@ -31,9 +32,11 @@ def pk(self):
 
 
 class TestAdmin(TestOrganizationMixin, TestCase):
-    """
+    '''
```
nemesisdesign, line 35: do not change this please

runtests.py
```diff
@@ -4,13 +4,13 @@
...
 
-if __name__ == "__main__":
+if __name__ == '__main__':
     from django.core.management import execute_from_command_line
 
     args = sys.argv
-    args.insert(1, "test")
-    args.insert(2, "openwisp_notifications")
+    args.insert(1, 'test')
+    args.insert(2, 'openwisp_notifications')
```
nemesisdesign, line 15: I think the reason is here. Look at the runtests.py script in the modules which implement swappable (eg: firmware-upgrader)

tests/openwisp2/sample_notifications/migrations/0001_initial.py
```diff
@@ -0,0 +1,173 @@
...
+                        null=True,
+                        on_delete=django.db.models.deletion.CASCADE,
+                        related_name='notify_target',
+                        to='contenttypes.ContentType',
+                    ),
+                ),
+                (
+                    'type',
+                    models.CharField(
+                        choices=NOTIFICATION_CHOICES, max_length=30, null=True,
```
nemesisdesign, line 162: :+1:

tests/openwisp2/sample_notifications/templates/admin/base.html
```diff
@@ -0,0 +1,11 @@
+{% extends "admin/base.html" %}
```
nemesisdesign, line 1: Is this really needed? I would have expected this template to be found by the DepdendencyLoader in openwisp-utils.

If this template is optional and is here only for example purposes, please state it both here (with a django template comment) and in the README.

pandafy: The tests fails without it. As I mentioned on IM: 

> The base template we use for admin site has URL for `openwisp_notifications`. 


Some tests in `test_admin.py` fails if this is not present, since Django is unable to resolve that URL.

pandafy:
I don't understand why `git` thinks I added `Settings` section in the `README` in this PR. :thinking:

PR #44, opened 2020-06-12
[models] Fixed bug for missing related object

### LINKED ISSUE #42: [enhancement] Make defining base_site.html optional
Presently for a derivative app, we need to define `admin/base_site.html` so that it can work properly. 
The main reason for this is presence of app label `openwisp_notifications` in `url` tag. 

A way to solve this will be to override `ModelAdmin` and pass app label as a variable.

### LINKED ISSUE #43: [bug] 'NoneType' object has no attribute 'name'
I deleted some metrics, then went to the notifications page and this exception was raised:

`AttributeError: 'NoneType' object has no attribute 'name'`

We need to make this code more resilient to failures:
https://github.com/openwisp/openwisp-notifications/blob/master/openwisp_notifications/base/models.py#L54-L59

Any item may be missing for some reason.

Try to replicate it by generating notifications related to a metric in openwisp-monitoring and then delete the metric, try deleting any of the generic relations

A few things need to be done:

- write tests for this bug, one test for each generic relation that can be missing
- avoid the operation to fail so drastically, make sure something is shown anyway but log warnings in case the data is not as expected

Suggestions:

- try to avoid accessing attribute names in the notification message string if possible, ideally data should be passed as a variable
- wrap the operation in a try except block and if there's an exception, log a warning and provide a default value

Full stack trace:

```
KeyError: 'display_list_message'
  File "django/db/models/options.py", line 581, in get_field
    return self.fields_map[field_name]
FieldDoesNotExist: Notification has no field named 'display_list_message'
  File "django/contrib/admin/utils.py", line 262, in lookup_field
    f = _get_non_gfk_field(opts, name)
  File "django/contrib/admin/utils.py", line 293, in _get_non_gfk_field
    field = opts.get_field(name)
  File "django/db/models/options.py", line 583, in get_field
    raise FieldDoesNotExist("%s has no field named '%s'" % (self.object_name, field_name))
AttributeError: 'NoneType' object has no attribute 'name'
  File "django/core/handlers/exception.py", line 34, in inner
    response = get_response(request)
  File "django/core/handlers/base.py", line 145, in _get_response
    response = self.process_exception_by_middleware(e, request)
  File "django/core/handlers/base.py", line 143, in _get_response
    response = response.render()
  File "django/template/response.py", line 105, in render
    self.content = self.rendered_content
  File "django/template/response.py", line 83, in rendered_content
    return template.render(context, self._request)
  File "django/template/backends/django.py", line 61, in render
    return self.template.render(context)
  File "django/template/base.py", line 171, in render
    return self._render(context)
  File "django/template/base.py", line 163, in _render
    return self.nodelist.render(context)
  File "django/template/base.py", line 936, in render
    bit = node.render_annotated(context)
  File "django/template/base.py", line 903, in render_annotated
    return self.render(context)
  File "django/template/loader_tags.py", line 150, in render
    return compiled_parent._render(context)
  File "django/template/base.py", line 163, in _render
    return self.nodelist.render(context)
  File "django/template/base.py", line 936, in render
    bit = node.render_annotated(context)
  File "django/template/base.py", line 903, in render_annotated
    return self.render(context)
  File "django/template/loader_tags.py", line 150, in render
    return compiled_parent._render(context)
  File "django/template/base.py", line 163, in _render
    return self.nodelist.render(context)
  File "django/template/base.py", line 936, in render
    bit = node.render_annotated(context)
  File "django/template/base.py", line 903, in render_annotated
    return self.render(context)
  File "django/template/loader_tags.py", line 150, in render
    return compiled_parent._render(context)
  File "django/template/base.py", line 163, in _render
    return self.nodelist.render(context)
  File "django/template/base.py", line 936, in render
    bit = node.render_annotated(context)
  File "django/template/base.py", line 903, in render_annotated
    return self.render(context)
  File "django/template/loader_tags.py", line 62, in render
    result = block.nodelist.render(context)
  File "django/template/base.py", line 936, in render
    bit = node.render_annotated(context)
  File "django/template/base.py", line 903, in render_annotated
    return self.render(context)
  File "django/template/loader_tags.py", line 62, in render
    result = block.nodelist.render(context)
  File "django/template/base.py", line 936, in render
    bit = node.render_annotated(context)
  File "django/template/base.py", line 903, in render_annotated
    return self.render(context)
  File "django/contrib/admin/templatetags/base.py", line 33, in render
    return super().render(context)
  File "django/template/library.py", line 214, in render
    _dict = self.func(*resolved_args, **resolved_kwargs)
  File "django/contrib/admin/templatetags/admin_list.py", line 342, in result_list
    'results': list(results(cl)),
  File "django/contrib/admin/templatetags/admin_list.py", line 318, in results
    yield ResultList(None, items_for_result(cl, res, None))
  File "django/contrib/admin/templatetags/admin_list.py", line 309, in __init__
    super().__init__(*items)
  File "django/contrib/admin/templatetags/admin_list.py", line 232, in items_for_result
    f, attr, value = lookup_field(field_name, result, cl.model_admin)
  File "django/contrib/admin/utils.py", line 271, in lookup_field
    value = attr(obj)
  File "openwisp_notifications/admin.py", line 116, in display_list_message
    return strip_tags(obj.message)
  File "django/utils/functional.py", line 48, in __get__
    res = instance.__dict__[self.name] = self.func(instance)
  File "openwisp_notifications/base/models.py", line 55, in message
    md_text = config['message'].format(notification=self)
```

pandafy:
Closes #43 

Caveats:

1. Receiver is triggered for all objects and makes two queries to database each time.
2. If number of notifications is huge, than this can take time. 

I thought of offloading this work to a celery worker, but it will not be possible as related objects may  already get deleted even before celery worker can complete this task. And we will get an error at assigning description. 

If we strictly opt for deleting notifications, that task can be offloaded to celery workers easily.

We are preserving information by not deleting notifications and leaving it to user when to delete such notifications. 

Hnet-image

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -74,6 +74,40 @@ Setup (integrate into an existing Django project)
         'openwisp_notifications',
      ]
 
+Configure caching (you may use a different cache storage if you want):
+
+.. code-block:: python
+
+    CACHES = {
+        'default': {
+            'BACKEND': 'django_redis.cache.RedisCache',
+            'LOCATION': 'redis://localhost/5',
+            'OPTIONS': {
+                'CLIENT_CLASS': 'django_redis.client.DefaultClient',
+            }
+        }
+    }
+
+    SESSION_ENGINE = 'django.contrib.sessions.backends.cache'
+    SESSION_CACHE_ALIAS = 'default'
+
+Configure celery:
+
+.. code-block:: python
+
+    # here we show how to configure celery with Redis but you can
+    # use other brokers if you want, consult the celery docs
+    CELERY_BROKER_URL = 'redis://localhost/1'
+
+    INSTALLED_APPS.append('djcelery_email')
+    EMAIL_BACKEND = 'djcelery_email.backends.CeleryEmailBackend'
+
+If you decide to use redis (as shown in these examples), install the requierd python packages:
+
+.. code-block:: shell
+
+    pip install redis django-redis
+
 Sending notifications
 ---------------------
 
@@ -351,6 +385,14 @@ publicly accessible from the internet. Otherwise, the logo may not be displayed 
 Installing for development
 --------------------------
 
+We use Redis as celery broker (you can use a different broker if you want).
+The recommended way for development is running it using Docker so you will need to
+`install docker and docker-compose <https://docs.docker.com/engine/install/>`_ beforehand.
+
+In case you prefer not to use Docker you can
+`install Redis from your repositories <https://redis.io/download>`_, but keep in mind that
+the version packaged by your distribution may be different.
+
 Install SQLite:
 
 .. code-block:: shell
@@ -371,6 +413,12 @@ Install test requirements:
 
     pip install -r requirements-test.txt
 
+Start Redis using docker-compose:
+
+.. code-block:: shell
+
+    docker-compose up -d
+
 Create a database:
 
 .. code-block:: shell
@@ -386,6 +434,13 @@ Launch the development server:
     ./manage.py runserver
 
 You can access the admin interface at http://127.0.0.1:8000/admin/.
+
+Run celery  worker (separate terminal window is needed):
+
+.. code-block:: shell
+
+    # (cd tests)
+    celery -A openwisp2 worker -l info
 
 Run tests with:
 
@@ -623,7 +678,26 @@ file in the test project.
 For more information about URL configuration in django, please refer to the
 `"URL dispatcher" section in the django documentation <https://docs.djangoproject.com/en/dev/topics/http/urls/>`_.
 
-12. Register Template Tags
+12. Create celery.py
+~~~~~~~~~~~~~~~~~~~~
+
+Please refer to the `celery.py <https://github.com/openwisp/openwisp-notifications/blob/master/tests/openwisp2/celery.py>`_
+file in the test project.
+
+For more information about the usage of celery in django, please refer to the
+`"First steps with Django" section in the celery documentation <https://docs.celeryproject.org/en/master/django/first-steps-with-django.html>`_.
+
+13. Import Celery Tasks
+~~~~~~~~~~~~~~~~~~~~~~~
+
+Add the following in your settings.py to import celery tasks from ``openwisp_notifications`` app.
+
+.. code-block:: python
+
+    CELERY_IMPORTS = ('openwisp_notifications.tasks',)
+
+
+14. Register Template Tags
 ~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 If you need to use template tags of *openwisp_notifications*, you will need to register as the, shown in
@@ -633,7 +707,7 @@ If you need to use template tags of *openwisp_notifications*, you will need to r
 For more information about template tags in django, please refer to the
 `"Custom template tags and filters" section in the django documentation <https://docs.djangoproject.com/en/dev/topics/http/urls/>`_.
 
-13. Add Base Template for Admin
+15. Add Base Template for Admin
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 Please refer to the `"templates/admin/base.html" in sample_notifications
@@ -643,7 +717,7 @@ For more information about customizing admin templates in django, please refer t
 `"Overriding admin templates" section in the django documentation
 <https://docs.djangoproject.com/en/3.0/ref/contrib/admin/#overriding-admin-templates>`_.
 
-14. Import the automated tests
+16. Import the automated tests
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 When developing a custom application based on this module, it's a good idea to import and run the base tests
--- a/README.rst
+++ b/README.rst
@@ -81,7 +81,7 @@ Configure caching (you may use a different cache storage if you want):
     CACHES = {
         'default': {
             'BACKEND': 'django_redis.cache.RedisCache',
-            'LOCATION': 'redis://localhost/5',
+            'LOCATION': 'redis://localhost/0',
             'OPTIONS': {
                 'CLIENT_CLASS': 'django_redis.client.DefaultClient',
             }
@@ -99,10 +99,7 @@ Configure celery:
     # use other brokers if you want, consult the celery docs
     CELERY_BROKER_URL = 'redis://localhost/1'
 
-    INSTALLED_APPS.append('djcelery_email')
-    EMAIL_BACKEND = 'djcelery_email.backends.CeleryEmailBackend'
-
-If you decide to use redis (as shown in these examples), install the requierd python packages:
+If you decide to use redis (as shown in these examples), make sure the python dependencies are installed in your system:
 
 .. code-block:: shell
 
--- a/openwisp_notifications/base/models.py
+++ b/openwisp_notifications/base/models.py
@@ -1,3 +1,5 @@
+import logging
+
 from django.conf import settings
 from django.contrib.sites.models import Site
 from django.core.cache import cache
@@ -12,9 +14,11 @@ from openwisp_notifications.types import (
     NOTIFICATION_CHOICES,
     get_notification_configuration,
 )
-from openwisp_notifications.utils import _get_object_link
+from openwisp_notifications.utils import NotificationException, _get_object_link
 
 from openwisp_utils.base import TimeStampedEditableModel, UUIDModel
+
+logger = logging.getLogger(__name__)
 
 
 class AbstractNotification(UUIDModel, BaseNotifcation):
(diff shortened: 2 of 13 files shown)


pandafy:
Tested with master branch of openwisp-monitoring

Hnet-image (1)

But the list display becomes a bit weird. 
Screenshot from 2020-06-12 18-25-48

pandafy:
@NoumbissiValere how about now? I didn't wanted to do string manipulation like below which makes code appear hacky. 
```
 message = obj.message.partition('<p><em>')[0]
```
Screenshot from 2020-06-15 10-23-27
Screenshot from 2020-06-15 10-23-19

NoumbissiValere:
It looks good. I will get it locally and test this evening

openwisp_notifications/handlers.py
```diff
@@ -158,3 +158,32 @@ def send_email_notification(sender, instance, created, **kwargs):
 )
 def clear_notification_cache(sender, instance, **kwargs):
     Notification.invalidate_cache(instance.recipient)
+
+
+@receiver(pre_delete, dispatch_uid='notification_related_object_deleted')
+def notification_related_object_deleted(sender, instance, **kwargs):
+    instance_id = getattr(instance, 'id', None)
```
nemesisdesign, line 165: use `pk` instead of `id` which is more reliable.

openwisp_notifications/handlers.py
```diff
@@ -158,3 +158,32 @@ def send_email_notification(sender, instance, created, **kwargs):
...
+        Q(actor_content_type=instance_content_type)
+        | Q(action_object_content_type=instance_content_type)
+        | Q(target_content_type=instance_content_type)
+    )
+    where = where & (
+        Q(actor_object_id=instance_id)
+        | Q(action_object_object_id=instance_id)
+        | Q(target_object_id=instance_id)
+    )
+    notifications = Notification.objects.filter(where)
```
nemesisdesign, line 177: Do you see any problem in just deleting the notifications?
Why do the extra work?
I mean, if a metric is deleted, we can't show notification related to it.
Same for devices or anything else the metric can be related to.

This would then become `Notification.objects.delete(where)` which is far more efficient.

Moreover, this logic could be easily executed in a celery background task, which would help us to keep response times under control.

requirements-test.txt
```diff
@@ -1,2 +1,4 @@
 coveralls
+django-redis
 openwisp-utils[qa]>=0.4.1
+redis
```
NoumbissiValere, line 4: I think this should move to the `requirements.txt` file as celery depends on it in a working environment

requirements-test.txt
```diff
@@ -1,2 +1,4 @@
 coveralls
+django-redis
```
NoumbissiValere, line 2: Same here

requirements-test.txt
pandafy: But we ask users to choose their own celery broker. Adding `redis` to `requirements.txt` will somehow bind them onto using Redis. That's why I added these in `requirements-test.txt` as it has been done in [openwisp-monitoring](https://github.com/openwisp/openwisp-monitoring/blob/master/requirements-test.txt).

NoumbissiValere: Yes @pandafy. That's a nice way of thinking. But let's know that we need to have default broker with which celery will work with when this code gets out there when deployed. The `requirements-test.txt` is not run in production. So while running your `requirements.txt` you should get everything you need to get your module up and running without any configuration. The tech gig will have the experience to get in and change their broker. You've done well by including the settings in the readme.

pandafy: Thank you @NoumbissiValere , I have made those changes. :smile:

NoumbissiValere:
Please also add a test case for this

pandafy:
Tested in openwisp-notifications for large number(100) of notifications.
Hnet-image (6)

Tested with Test App included in sample_notifications
Hnet-image (5)

Tested with master branch of openwisp-monitoring, when metric is deleted. 
Hnet-image (4)

requirements-test.txt
nemesisdesign: I'm sorry I was not fast to reply to this, but @NoumbissiValere you are not right regarding to this, @pandafy was following what we already do in other modules.
The fact that celery uses redis, is a deployment choice, and not an application choice. Celery can be deployed using many different brokers (RabbitMQ, memcached, database) and we should not force the usage of redis.
Please keep in mind for the next time.

tests/openwisp2/settings.py
```diff
@@ -118,3 +162,5 @@
     TEMPLATES[0]['DIRS'].insert(
         0, os.path.join(BASE_DIR, 'sample_notifications', 'templates')
     )
+    # Celery auto detects tasks only from INSTALLED_APPS
+    CELERY_IMPORTS = ('openwisp_notifications.tasks',)
```
nemesisdesign, line 166: mention this in the documentation as it's done in [openwisp-monitoring](https://github.com/openwisp/openwisp-monitoring#import-celery-tasks)

tests/openwisp2/settings.py
```diff
@@ -98,6 +99,49 @@
     },
 ]
 
+CACHES = {
+    'default': {
+        'BACKEND': 'django_redis.cache.RedisCache',
+        'LOCATION': 'redis://localhost/0',
```
nemesisdesign, line 105: let's change this to `redis://localhost/5`, it allows us to run celery workers of different modules in development without them clashing (when all modules use 0, if you run the celery worker of the development environment of 2 different modules at the same time, they will try to load each other task, not finding them and returning errors)

requirements.txt
```diff
@@ -1,6 +1,9 @@
 django-notifications-hq>=1.6,<1.7
+django-celery-email>=3.0.0,<3.1
+django-redis
```
nemesisdesign, line 3: as said in the previous comment, move back to `requirements-test.txt`

requirements.txt
```diff
@@ -1,6 +1,9 @@
...
 openwisp-users>=0.2.0,<0.3.0
 # TODO: enable when openwisp-utils 0.5 is released
 # openwisp-utils[rest]~=0.5
+redis
```
nemesisdesign, line 7: same here

openwisp_notifications/base/models.py
```diff
@@ -51,25 +55,36 @@ def message(self):
...
-                site=Site.objects.get_current(), notification=self
-            )
+            try:
+                config = get_notification_configuration(self.type)
+                return config['email_subject'].format(
+                    site=Site.objects.get_current(), notification=self
+                )
+            except AttributeError as e:
+                logger.exception(e)
+                return "Attribute Error, check logs"
```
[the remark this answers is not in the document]
nemesisdesign: can't we just raise an exception here and catch it in the receiver which sends the email?
Returning this string is really weird and confusing!

requirements-test.txt
NoumbissiValere: Thank you @nemesisdesign and sorry @pandafy :smile:

openwisp_notifications/base/models.py
```diff
@@ -51,12 +55,19 @@ def message(self):
...
+                else:
+                    md_text = render_to_string(
+                        config['message_template'], context=dict(notification=self)
+                    ).strip()
+            except AttributeError as e:
+                logger.error(e)
+                md_text = (
+                    'Attribute Error while generating notification message.'
+                    'Check logs for further information.'
+                )
```
nemesisdesign, line 70: I don't know, this feels relaly wrong to me.

The best thing would be to skip the notification entirely, but I guess we cannot do that in the admin.

However, I believe we can do this once we use the JS widget to show notifications.

I think the best course of action is the following. For now, just make the message more generic: 
`Error while generating notification message, notification data may have been deleted`.
I don't think we should advise users to check logs, because it's entirely possible they may not have access to the application logs.

Later on, once we have the notification widget and we stop using the django admin, we can raise an exception (a custom one we create for this purpose), catching this exception in the API and websocket when iterating on the notification list, in case the exception is catched, the rendering of the notification is just skipped.

Will create an issue for this.

requirements.txt
```diff
@@ -1,3 +1,4 @@
+django-celery-email>=3.0.0,<3.1
```
nemesisdesign: @pandafy this shouldn't be here as it's not a direct dependency of openwisp-notifications. I'm going to remove manually.

README.rst
```diff
@@ -74,6 +74,40 @@ Setup (integrate into an existing Django project)
...
+Configure celery:
+
+.. code-block:: python
+
+    # here we show how to configure celery with Redis but you can
+    # use other brokers if you want, consult the celery docs
+    CELERY_BROKER_URL = 'redis://localhost/1'
+
+    INSTALLED_APPS.append('djcelery_email')
+    EMAIL_BACKEND = 'djcelery_email.backends.CeleryEmailBackend'
```
nemesisdesign, line 103: I don't know why you're mentioning this, it's not needed in what we're doing, I already told you, this is a setting that is performed during deployment and not at application level.

pandafy: I thought this would be helpful for users. Since we deals with emails here, I thought this would be appropriate place to mention this. 

Sorry for the commit message.

nemesisdesign: You may add this suggestion in a different patch, but ensure it's mentioned in a different block of text, eg:

```
# pseudo docs
If you want to configure email sending to happen in the background, do the following:

- pip install ...
- edit settings as follows...
```

PR #59, opened 2020-07-14
[fix] Reduced database queries for notifications #47

### LINKED ISSUE #47: Too many queries
@pandafy have you enabled the debug toolbar?
See here: http://openwisp.io/docs/developer/hacking-openwisp-python-django.html#django-debug-toolbar

Have you noticed the amount of queries?
Before we were storing the message but now we're generating it each time which generates far too many queries, so we should find a solution to this using caching (and we should handle cache invalidation).

one possible way of doing is the following :

- cache the `actor` object and try to retrieve it from the cache before querying
- same for `target`

Once the problem is fixed, we should add assertNumQueries in tests to make sure the amount of queries is kept under control.

I will arrange this in the board so it has less priority over the API and the JS widget.

pandafy:
Added caching for related objects to limit database queries.
Closes #47 

For 18 notifications, 58 database queries were being made each time earlier on Notification Admin. After caching they are down to 4 . :smile: 

**Before**
Screenshot from 2020-07-14 13-34-48

**After**
Screenshot from 2020-07-14 13-35-23

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -254,7 +254,7 @@ Additional ``notify`` keyword arguments
 |                     | not be added to the email text.                                             |
 +---------------------+-----------------------------------------------------------------------------+
 |       ``type``      | Set values of other parameters based on predefined setting                  |
-|                     | ``OPENWISP_NOTIFICATION_TYPES``                                             |
+|                     | ``OPENWISP_NOTIFICATIONS_TYPES``                                            |
 |                     |                                                                             |
 |                     | Defaults to **None** meaning you need to provide other arguments.           |
 +---------------------+-----------------------------------------------------------------------------+
@@ -399,8 +399,8 @@ registered.
 Settings
 --------
 
-``OPENWISP_NOTIFICATION_HTML_EMAIL``
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+``OPENWISP_NOTIFICATIONS_HTML_EMAIL``
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 +-----------+------------+
 |   type    |  ``bool``  |
@@ -412,8 +412,8 @@ If ``True``, attaches markdown rendered HTML of notification message in email no
 If ``False``, HTML rendering of notification message will be disabled and a plain
 text email is sent.
 
-``OPENWISP_NOTIFICATION_EMAIL_TEMPLATE``
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+``OPENWISP_NOTIFICATIONS_EMAIL_TEMPLATE``
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 +-----------+--------------------------------------------------+
 |   type    |  ``str``                                         |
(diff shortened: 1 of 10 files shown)


openwisp_notifications/base/models.py
```diff
@@ -86,6 +119,28 @@ def email_subject(self):
         else:
             return self.message
 
+    def _related_object(self, field):
+        obj_id = getattr(self, f'{field}_object_id')
+        if obj_id:
```
nemesisdesign, line 124: Please use this form to avoid nesting the rest of the code and hence improving readability:
```python
if not obj_id:
    return
```

openwisp_notifications/base/models.py
```diff
@@ -86,6 +119,28 @@ def email_subject(self):
...
+            cache_key = self._cache_key(obj_id)
+            obj = cache.get(cache_key)
+            if not obj:
+                obj = getattr(self, f'_{field}')
+                cache.set(cache_key, obj)
```
nemesisdesign, line 129: add a timeout and make it configurable, I'd set it to 5 days by default

README.rst
```diff
@@ -434,6 +444,19 @@ This setting takes the URL of the logo to be displayed on email notification.
 **Note**: Provide a URL which points to the logo on your own web server. Ensure that the URL provided is
 publicly accessible from the internet. Otherwise, the logo may not be displayed in email.
 
+OPENWISP_NOTIFICATION_CACHE_TIMEOUT
```
nemesisdesign, line 447: `OPENWISP_NOTIFICATION_CACHE_TIMEOUT` > `OPENWISP_NOTIFICATIONS_CACHE_TIMEOUT`

.travis.yml
```diff
@@ -24,6 +24,7 @@ branches:
   only:
     - master
     - dev
+    - issues/12-websockets
```
nemesisdesign, line 27: you can remove this

openwisp_notifications/handlers.py
```diff
@@ -158,13 +157,23 @@ def send_email_notification(sender, instance, created, **kwargs):
     instance.save()
 
 
+@receiver(post_save, dispatch_uid='notification_related_object_cache')
+@receiver(post_delete, dispatch_uid='notification_related_object_cache_deleted')
```
nemesisdesign, line 161: this is worrying to me, if we don't specify a sender, this invalidation will be triggered constantly, which would make all this work you're doing useless.

Maybe we can just rely on the timeout and decrease it to 2 days and avoid doing this at all.

openwisp_notifications/handlers.py
```diff
@@ -158,13 +157,23 @@ def send_email_notification(sender, instance, created, **kwargs):
...
+def invalidate_notification_related_object_cache(sender, instance, **kwargs):
+    if kwargs.get('created', False):
+        return
+    instance_id = getattr(instance, 'pk', None)
+    if instance_id:
+        Notification.invalidate_related_object_cache(instance_id)
```
nemesisdesign, line 167: here you're invalidating based on instance_id but not supplying any info on the content type which doesn't look right

openwisp_notifications/base/models.py
```diff
@@ -86,6 +120,33 @@ def email_subject(self):
         else:
             return self.message
 
+    def _related_object(self, field):
+        obj_id = getattr(self, f'{field}_object_id')
+        if not obj_id:
+            return
+        cache_key = self._cache_key(obj_id)
```
nemesisdesign, line 127: please take in consideration the django content type as well when calculating the cache key, you could try to use the content type id to avoid generating the query to fetch the content type object

openwisp_notifications/settings.py
```diff
@@ -27,6 +27,10 @@
     '/static/openwisp_notifications/audio/notification_bell.mp3',
 )
 
+OPENWISP_NOTIFICATIONS_CACHE_TIMEOUT = getattr(
+    settings, 'OPENWISP_NOTIFICATIONS_CACHE_TIMEOUT', 5 * 24 * 60 * 60
```
nemesisdesign, line 31: seems you forgot to set 2 days here

PR #69, opened 2020-07-22
[feature] Added disabling of notifications by type, organization and medium

### LINKED ISSUE #5: [feature] Notification settings: allow users to disable notifications by type, organization, medium
**Goals**
- [x] Allow user to disable notifications by notification type. 
- [x] Allow possibility to disable only email notifications or web notifications.
- [x] Allow disabling notification for an organization.

**Project Idea**
> Allow users to disable notifications by notification types, right now it's only possible to enable or disable the notifications globally.
For each notification type and organization the user should be able to disable either all notifications or only the email notification.
Example: Receive notifications for organization "default"?
Two checkboxes: Web and Email; if email is checked also web is checked automatically (because the notification depends on the basic notification being created).
Such a solution would give us the possibility to implement mobile push notifications in the future by adding a new checkbox for mobile notifications.

> Allow users to disable notifications for specific objects, either permanently or until a specific date, with predefined set of durations: permanently, 1 day, 3 days, 1 week, 1 month, custom date. In the case the disabling is temporary, the setting should be deleted automatically after expiration by a simple celery task which is run periodically with celery beat.

pandafy:
Closes #5


openwisp_notifications/admin.py
```diff
@@ -6,11 +6,36 @@
...
 NotificationUser = load_model('NotificationUser')
+NotificationSettings = load_model('NotificationSetting')
 
 
 class NotificationUserInline(AlwaysHasChangedMixin, admin.StackedInline):
     model = NotificationUser
     fields = ['receive', 'email']
 
 
+class NotificationSettingAdmin(AlwaysHasChangedMixin, admin.TabularInline):
```
nemesisdesign, line 17: `NotificationSettingAdmin` > `NotificationSettingsInline`

openwisp_notifications/tasks.py
```diff
@@ -25,3 +31,66 @@ def delete_obsolete_notifications(instance_app_label, instance_model, instance_i
...
+# Following tasks add notification settings in database
+
+
+@shared_task
+def ns_register_notification_type(notification_type):
+    """
+    Adds notification setting for all users who are memeber of an
+    organization when a new notification type is registered.
+    """
+    for user in User.objects.all():
```
nemesisdesign, line 45: use `User.objects.iterator()`

openwisp_notifications/tasks.py
```diff
@@ -25,3 +31,66 @@ def delete_obsolete_notifications(instance_app_label, instance_model, instance_i
...
+
+@shared_task
+def ns_organization_user_deleted(instance_user_id, instance_org_id):
+    """
+    Deletes notification settings for all notification types when
+    an organization user is deleted.
+    """
+    NotificationSetting.objects.filter(
+        user_id=instance_user_id, organization_id=instance_org_id
+    ).delete()
```
nemesisdesign, line 96: are you sure this is necessary? If the `organization` field is set, the SQL DELETE should cascade.

openwisp_notifications/tasks.py
```diff
@@ -25,3 +31,66 @@ def delete_obsolete_notifications(instance_app_label, instance_model, instance_i
...
+            try:
+                NotificationSetting.objects.create(
+                    user_id=user.id, organization_id=org_user, type=notification_type,
+                )
+            except IntegrityError:
+                continue
+
+
+@shared_task
+def ns_unregister_notification_type(notification_type):
```
nemesisdesign, line 66: I am not loving these names, what about names like `create_settings_for_` and `delete_settings_for_`?

openwisp_notifications/tasks.py
pandafy: It is a foreign key to `Organization` not `OrganizationUser`.

openwisp_notifications/tasks.py
pandafy: Those were looking long, `ns` is short for `notificationsetting`

pandafy:
I added a migrate command in `travis.yml` and build does not fail for sample app. The culprit is a notification which is being registered in ready function of `SampleNotificationsConfig`. This triggered task for creating notification setting, but migrations are not run yet, therefore we were getting error. Can we keep the migration command?

I tried using `post_migrate` signal for registering notifications, butt then we can not register notifications from anywhere in code.
We will be restricted to `apps.py` and it will require us to refactor current tests.

nemesisdesign:
> I added a migrate command in `travis.yml` and build does not fail for sample app.

Does not fail or does fail?

pandafy:
> Does not fail or does fail?

The build is not failing when we run migrations on database before running tests. And I was wrong, the error is still happening even when migrations are run before tests. The difference is the error are popping out in celery worker and not in main thread. 

The only thing I don't understand that we do almost identical work in [device/apps.py](https://github.com/openwisp/openwisp-monitoring/blob/f2733dfd288b29b67633f05fdeccd1d2312fe687/openwisp_monitoring/device/apps.py#L24) and [monitoring/apps.py](https://github.com/openwisp/openwisp-monitoring/blob/master/openwisp_monitoring/monitoring/apps.py#L22) in openwisp-monitoring and there I don't see any error even in celery worker. Is there anything I am missing here? 

> what is default type?

It is the default notification type which comes with openwisp-notifications. We can un-register it in openwisp-monitoring if we don't want it.

pandafy:
I have tested that migrations can go back and forth swiftly both on openwisp-notifications and openwisp-monitoring. 
I have also manual tested all features on openwisp-monitoring, please see following
- [Screenshots of available settings in openwisp-monitoring](https://github.com/openwisp/openwisp-monitoring/pull/171#issuecomment-671075713)
- [A GIF of working functionalities in openwisp-monitoring](https://user-images.githubusercontent.com/32094433/89739293-2b560600-da9d-11ea-9bdb-ce92c89dc3be.gif)

i will update screenshot of live API documentation after #102 is merged, because endpoints conflicts.

pandafy:
Thanks for the review. I will make these changes. :smile: 
> * I pulled the code, switched to the branch, ran migrations and ran the dev server, but I didn't see the rows being created for the default notification type

Was celery worker running in background? :sweat_smile:

nemesisdesign:
> > * I pulled the code, switched to the branch, ran migrations and ran the dev server, but I didn't see the rows being created for the default notification type
> 
> Was celery worker running in background?

It wasn't, after running it they appeared, thanks :+1:

pandafy:
I implement the requested features and tested them on openwisp-monitoring. 

Hnet-image (6)

In the GIF, I have set `web_notification` option of `threshold_recovery` notification type to `False`. Thus no notification was received before. But after defining values by user, it was working as expected.

openwisp_notifications/base/models.py
```diff
@@ -178,26 +180,71 @@ def redirect_view_url(self):
...
     def save(self, *args, **kwargs):
-        if not self.receive:
-            self.email = self.receive
-        return super(AbstractNotificationUser, self).save(*args, **kwargs)
+        if not self.web_notification:
+            self.email = self.web_notification
+        return super().save(*args, **kwargs)
+
+    def full_clean(self, *args, **kwargs):
+        notification_type_config = get_notification_configuration(self.type)
```
nemesisdesign, line 233: I'd create a property for this, eg:

`self.config`

openwisp_notifications/base/models.py
```diff
@@ -178,26 +180,71 @@ def redirect_view_url(self):
...
+            self.email = None
+        if self.web == notification_type_config['web_notification']:
+            self.web = None
+        return super().full_clean(*args, **kwargs)
+
+    @cached_property
+    def email_notification(self):
+        if self.email is not None:
+            return self.email
+        return get_notification_configuration(self.type).get('email_notification')
```
nemesisdesign, line 244: this then could be shortened to 
```python
return self.email or self.config.get('email_notification')
```

README.rst
```diff
@@ -690,6 +723,27 @@ Delete a notification
 
     DELETE /api/v1/notification/{pk}/
 
+List user's notification setting
+################################
+
+.. code-block:: text
+
+    GET /api/v1/notifications/setting/
```
nemesisdesign, line 731: `notifications` > `notification`.

I also think it would be more explicit to name it user-setting

openwisp_notifications/base/models.py
pandafy: `self.config` can be misleading. I have used `self.notification_type_config` instead.

openwisp_notifications/base/models.py
pandafy: I don't think we can use `or` here because if will always return `True` even when other is `False`.
```
>>> False or True
True
```

I have added the property though.

pandafy:
> * if the user is not superuser and member of only one org, maybe we can avoid creating the setting for that org? I think this would result in the UI being simpler.

This way user won't be able to configure their notification. What if the user just want to receive web notification?

pandafy:
> * pay attention to the number of queries generated in the user page, it's far too high now (!), try to use `prefetch_related` in the queryset of the inline

It was inline form which was making queries for populating choice field for organizations which was resulting in so many queries. Fixed it. 

Screenshot_20200813_204033

openwisp_notifications/admin.py
```diff
@@ -1,16 +1,62 @@
...
-NotificationUser = load_model('NotificationUser')
+NotificationSetting = load_model('NotificationSetting')
 
 
-class NotificationUserInline(AlwaysHasChangedMixin, admin.StackedInline):
-    model = NotificationUser
-    fields = ['receive', 'email']
+@admin.register(NotificationSetting)
+class NotificationSettingAdmin(
+    NotificationSettingAdminMixin, admin.ModelAdmin,
```
nemesisdesign, line 14: why did you add this? What can this do that we can't already do from the user admin?

openwisp_notifications/admin.py
```diff
@@ -1,16 +1,62 @@
...
+        finally:
+            return user_owns_obj or request.user.is_superuser
+
+
+class NotificationSettingInline(
+    NotificationSettingAdminMixin, AlwaysHasChangedMixin, admin.TabularInline
+):
+    model = NotificationSetting
+    extra = 0
+
```
nemesisdesign, line 57: I can see a decrease of queries if I add:

```python
def get_queryset(self, request):
        return super().get_queryset(request).select_related('organization')
```

Can you see the same?

openwisp_notifications/base/models.py
```diff
@@ -178,26 +180,74 @@ def redirect_view_url(self):
...
+
+    def full_clean(self, *args, **kwargs):
+        if self.email == self.notification_type_config['email_notification']:
+            self.email = None
+        if self.web == self.notification_type_config['web_notification']:
+            self.web = None
+        return super().full_clean(*args, **kwargs)
+
+    @cached_property
+    def notification_type_config(self):
```
nemesisdesign, line 240: what do you think about `type_config`?

openwisp_notifications/handlers.py
```diff
@@ -50,8 +54,22 @@ def notify_handler(**kwargs):
...
+        # Create notification for users who have opted for receiving notifications.
+        # For users who have not configured web_notifications,
+        # use default from notification type
+        web_notification = Q(notificationsetting__web=True)
+        if notification_template['web_notification']:
+            web_notification |= Q(notificationsetting__web=None)
+
+        notification_setting = web_notification & Q(
+            notificationsetting__type=notification_type,
+            notificationsetting__organization_id=target_org,
```
nemesisdesign, line 69: since we're querying for these fields, I see that we don't have an index on the type field, can you please investigate what's the best way to add a dedicated index to speed up this query? Either a single index or a composite index, but the query should be fast and be able to scale also when the amount of users in the system is high.

openwisp_notifications/admin.py
pandafy: Operators can not view `User` admin, for them I had to add this. 
This also simplifies editing notification settings by superuser as they can filter via notification types and organization and edit notification settings for users accordingly.  We can hide this for `Superusers` and `Administrators` if you like.

openwisp_notifications/base/models.py
pandafy: :+1:

openwisp_notifications/admin.py
nemesisdesign: I want to avoid having to add more items to the admin index page, because on installations which use many modules it becomes huge, at some point all those links will have to go in the menu because the index will give way to a general dashboard, but also the menu cannot be huge or users will be overwhelmed.

I suggest the following:

- we can just remove this now
- we can open an issue to add a dedicated admin view to allow the current logged in user to change their notification settings, we can then add a link like "change notification preferences" somewhere in the notification JS widget, this would be great for UX

openwisp_notifications/handlers.py
pandafy: I created 100 superusers and 100 organizations with 10 notification types. Following are graphs for 20 cold start(django caches queryset results). Please have a look at this [spreadsheet containing all the data](https://docs.google.com/spreadsheets/d/1N-Z3mQldWJvi7yBgMX7vce2CBTMpNh8N0CA3mJA1q00/edit?usp=sharing) for more information. I think 'type' and 'organization' could be a good candidate, because there will certainly not be as many number of organizations as users in production systems. 


Look up for users (expected when creating notification)
Lookup for notification settings(expected in admin site) (1)

pandafy:
@nemesisdesign "All" organizations created a confusion, Actually the whole purpose for adding an "All" setting was to allow notifications for devices(target objects) without organizations. But since notifications are only sent to superusers if target organization is not present, I have removed the option of "All" from organization users.

README.rst
```diff
@@ -460,6 +474,25 @@ Then in the application code:
             error=str(error)
         )
 
+Notification Preference
```
nemesisdesign, line 477: Notification Preferences

pandafy:
> 1. NotificationSetting objects created for `OrganizationUser(is_admin=False)`

Fixed this one. added `user__is_staff` as well in the query. 

> 2. All vs org settings

I think `All` settings was poorly implemented or the was misleading. The goal of providing an all setting was to allow super users control notifications which does not belong to any `organization` i.e., a target object which does not belong to any organization. I did not provide an option to override settings because I did not aimed for it. Since, notifications which we are dealing now are created for `Device` object which has organization as required field, so it does not matter much to have control over such notifications. But, it will be a nice to have feature, we might need to think about accessibility for user and implement it accordingly. What do you think?

> 3. Checking vs Unchecking, WTF?

I should have seen this coming. Fixed it using JS. I have also retained the logic in models since these settings can also be modified from API.  
Hnet-image (8)


I also caught an edge case during fixing these. 
- When an organizationuser is modified, say it's `is_admin` or `organization` attribute is updated, 

Also fixed this!

PR #73, opened 2020-07-28
[enhancement] Added redirect view for marking notification as read

### LINKED ISSUE #20: [email] Mark notification as read when user opens link to related object from email.
Instead of providing link to notification, provide link to related object. On opening this link, this notification should be automatically marked as read.

### LINKED ISSUE #51: [refractor] Make related object links cached_property
We are generating links to related objects at multiple places.

1. At admin dashboard.
2. While rendering markdown message
3. While fetching details of notifications using API. 

 We can instead create a property for these links and use them at other places.

pandafy:
Closes #20
Closes #51

### DIFF
--- a/openwisp_notifications/base/models.py
+++ b/openwisp_notifications/base/models.py
@@ -66,19 +66,27 @@ class AbstractNotification(UUIDModel, BaseNotifcation):
         """
         cache.delete(cls.count_cache_key(user.pk))
 
+    @property
+    def actor_url(self):
+        return _get_object_link(self, field='actor', url_only=True, absolute_url=True)
+
+    @property
+    def action_url(self):
+        return _get_object_link(
+            self, field='action_object', url_only=True, absolute_url=True,
+        )
+
+    @property
+    def target_url(self):
+        return _get_object_link(self, field='target', url_only=True, absolute_url=True)
+
     @cached_property
     def message(self):
         if self.type:
             # setting links in notification object for message rendering
-            self.actor_link = _get_object_link(
-                self, field='actor', url_only=True, absolute_url=True
-            )
-            self.action_link = _get_object_link(
-                self, field='action_object', url_only=True, absolute_url=True,
-            )
-            self.target_link = _get_object_link(
-                self, field='target', url_only=True, absolute_url=True
-            )
+            self.actor_link = self.actor_url
+            self.action_link = self.action_url
+            self.target_link = self.target_url
 
             config = get_notification_configuration(self.type)
             try:
--- a/openwisp_notifications/base/models.py
+++ b/openwisp_notifications/base/models.py
@@ -6,6 +6,7 @@ from django.contrib.sites.models import Site
 from django.core.cache import cache
 from django.db import models
 from django.template.loader import render_to_string
+from django.urls import reverse
 from django.utils.functional import cached_property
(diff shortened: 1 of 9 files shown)


pandafy:
When user is already logged in.

Hnet-image (12)

When user is logged out.
Hnet-image (13)

openwisp_notifications/base/models.py
```diff
@@ -151,6 +153,18 @@ def action_object(self):
     def target(self):
         return self._related_object('target')
 
+    @property
+    def redirect_view_link(self):
+        return _get_absolute_url(
+            '{redirect_view}?target_url={target}'.format(
+                redirect_view=reverse(
+                    f'{self._meta.app_label}:notification_read_redirect',
```
nemesisdesign, line 161: app_label and URL namespace are two different concepts, do not mix them.
Let's use always `notifications` as URL namespace, see https://github.com/openwisp/openwisp-notifications/commit/5bb699dabf50a56bc370ec9262ceeb4da74201c0 to know why.

pandafy:
> 1. One link does not point to redirect URL

The target object URL which does not redirec,t is generated through notification message. If we add read-redirect link there, it will redirect for notifications in JS widget as well. 

> 2. Redirect URL not found.

From the notification email and screenshot that you have shared, it seems like this notification belongs to `operator` and you logged in with `admin` account. Shouldn't this be expected? Please correct me if I am wrong. :sweat_smile: 

> 3. URL format can be simplified and made more RESTful

I thought, we could save some processing time for user from building URL at time of redirecting, I will amend that.

nemesisdesign:
> > 1. One link does not point to redirect URL
> 
> The target object URL which does not redirec,t is generated through notification message. If we add read-redirect link there, it will redirect for notifications in JS widget as well.

We have to find a solution, what do you propose?

Having two links in the email, one which points to the redirect URL and one which doesn't, can potentially make the whole work useless, because we don't know whether users will click on one or the other, users may very well click mostly
Both links have to point to the redirect URL.

> > 1. Redirect URL not found.
> 
> From the notification email and screenshot that you have shared, it seems like this notification belongs to `operator` and you logged in with `admin` account. Shouldn't this be expected? Please correct me if I am wrong. 

Maybe you're right, I'll try again.

> > 1. URL format can be simplified and made more RESTful
> 
> I thought, we could save some processing time for user from building URL at time of redirecting, I will amend that.

Don't worry about this, the system will be able to handle this because users won't click on those links so often.

openwisp_notifications/tests/test_api.py
```diff
@@ -425,3 +432,45 @@ def test_obsolete_notifications_busy_worker(self, mocked_task):
...
+            self.client.logout()
+            url = reverse(
+                f'{self.url_namespace}:notification_read_redirect',
+                args=(notification.pk,),
+            )
+            response = self.client.get(url)
+            self.assertEqual(response.status_code, 302)
+            self.assertEqual(
+                response.url,
+                f'/admin/login/?next=/api/v1/notifications/{notification.pk}/redirect/',
```
nemesisdesign, line 475: would have been better to use `reverse()` for consistency.
It's also good to use `reverse()` for maintainability, although this form is more readable.
Not worth it to debate these minor details though :grin:

PR #97, opened 2020-08-06
[fix] Fixed UX issues of notification toast #94

### LINKED ISSUE #94: [ux] Problems found while testing toast notifications
Toasts are really great and very useful, they will allow us to do great things and inform the user in real time about what is going on in the network.

But we need to make the current implementation more robust.

- ~~Have one style for each level~~ (handled in https://github.com/openwisp/openwisp-notifications/pull/92/commits/07200acecae30362368da10d379ad6765b89c75f, PR #92)
- Keep the toast open longer (eg: 15 seconds?) so users won't miss them
- Add a close icon button to close it
- Mark as read on click or on close
- Allow having multiple toast notifications stacked, right now if I create two notifications, the last one cover the previous one, but as our usage of this tool increases, I am 100% sure in a busy system it's very likely two or more notifications can appear in a short time span when launching important changes to the network and we have to handle this well

pandafy:
Increased timeout for removing notifications from display
Added a close button for closing notification toast
Closing notification toast also flags that notification as read
Allowed notification toasts to stack instead of overlaying over each other

Closes #94


pandafy:
@nemesisdesign would you prefer making notification toast wrapper overflow hidden(or scrollable) as shown in first image or just keep them as in second image?
I personally like the idea of scrollable notification toasts. Let me know your thoughts, it will be a quick change. :smile: 

Screenshot from 2020-08-06 16-57-20
Screenshot from 2020-08-06 16-56-49

PR #103, opened 2020-08-08
[fix] Fixed notification toast

pandafy:
Clicking on close button, opened target object link.
Fixed it by stopping propagation of click event.
Continuation of #97


openwisp_notifications/management/commands/__init__.py
```diff
@@ -7,4 +7,4 @@
 class BaseCreateNotificationCommand(BaseCommand):
     def handle(self, *args, **kwargs):
         default_org = Organization.objects.first()
-        notify.send(sender=default_org, type='default')
+        notify.send(sender=default_org, type='default', target=default_org)
```
nemesisdesign, line 10: try to not mix these changes.. it makes the history harder to understand

PR #104, opened 2020-08-10
[fix] Fixed notification widget loader #96

### LINKED ISSUE #96: [ux] Show loading indicator while loading notifications
The current loading indicator is limited because it can be used only when the js widget is empty.

When scrolling to the bottom, there's no indication in the UI that new notifications are being loaded, this is not good for ux because the scrollbar suddenly gets bigger and it may not be entirely clear to the user what has happened, while a loading indicator would surely make this clearer.

One possibility could be to show the same loading indicator at the end of the scrollable box in an empty space, but I'm open for other ideas if you have any.

pandafy:
Fixed loader not shown while fetching more notifications.
Closes #96

### DIFF
--- a/openwisp_notifications/static/openwisp-notifications/css/notifications.css
+++ b/openwisp_notifications/static/openwisp-notifications/css/notifications.css
@@ -204,19 +204,13 @@
 }
 
 /* Loader */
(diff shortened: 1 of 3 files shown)


pandafy:
Hnet-image (4)

PR #108, opened 2020-08-11
[enhancement] Load notification on user interaction #100

### LINKED ISSUE #100: [ux] Do not load notifications without user interaction
I have noticed that at each page load the widget generates an API request to retrieve the notifications, but this is not necessary.
Notifications should be loaded only when the user clicks on the bell icon.

pandafy:
API call to fetch notification is only made when user interacts
with notification widget(clicks bell icon).
Closes #100


pandafy:
initialisation API call is only made when notification bell is clicked for the first time. I have also tested on openwisp-monitoring. 
Hnet-image (7)

PR #111, opened 2020-08-18
 [feature] Added Object Notification Setting #38

### LINKED ISSUE #38: [feature] Allow users to ignore notifications on a specific object
- [ ] Allow a user to ignore notifications for a specific object, permanently or temporarily (1 day, 3 days, 1 week, 1 month, custom duration)
- [ ] Use celery to enable disabled notifications after passage of the specified time
- [ ] Provide an API to activate this feature of an object (eg: `PUT /api/v1/notification/ignore/<content-type>/<object-id>/`)
- [ ] Provide an API to deactivate this feature for an object (eg: `DELETE /api/v1/notification/ignore/<content-type>/<object-id>/`)

**Project Idea**
>Allow users to disable notifications for specific objects, either permanently or until a specific date, with predefined set of durations: permanently, 1 day, 3 days, 1 week, 1 month, custom date. In the case the disabling is temporary, the setting should be deleted automatically after expiration by a simple celery task which is run periodically with celery beat.

pandafy:
- Added ObjectNotification model to add option to disable
  notification for a particular device
- Added API endpoints for ObjectNotification model

Closes #38


openwisp_notifications/base/models.py
```diff
@@ -250,3 +251,15 @@ def web_notification(self):
         if self.web is not None:
             return self.web
         return self.type_config.get('web_notification')
+
+
+class AbstractObjectNotification(UUIDModel):
```
nemesisdesign, line 256: from reading this code is not immediately clear what this model does.

Does it serve as ignore?
Does every object need this? Or is it added only when a user deactivated notifications for a particular object?

pandafy:
I have renamed `ObjectNotification` to `IgnoreObjectNotification`, I hope it is more descriptive. I have also added doc string to the abstract model to make it explicit what this model is for. 

I have used `button` elements for the drop down widget and tested accessibility from keyboard.
Hnet-image (20)

> ```css
> .ow-object-notification-option-container p {
>     margin: auto;
>     font-size: 12px;
>     padding: 8px 0;
>     outline: 0.5px solid rgba(0, 0, 0, 0.05);
> }
> ```

Instead of outline here, i went with border. While focusing though buttons, an outline appears which is of a darker colour(`#101010`) inherited from`user agent stylesheet` . If we overide this setting, it makes it almost impossible to see which element is in focus.  

> I clicked on Unsubscribe, the tooltip appearing on hover to the button has stayed "You are receiving notifications for this object."
Fixed it. :sweat_smile: 
Hnet-image (22)

pandafy:
That was a silly mistake. I have updated the code. 
Screenshot_20200827_185730

README.rst
```diff
@@ -575,6 +575,17 @@ by disabling both web and email option for a notification setting.
 then ``email_notification`` or ``web_notification`` option of concerned notification type will be used
 respectively.
 
+
```
nemesisdesign, line 578: let's avoid this extra blank line for consistency

openwisp_notifications/widgets.py
```diff
@@ -0,0 +1,39 @@
...
+class IgnoreObjectNotificationWidgetMedia:
+    extend = True
+    js = [
+        'admin/js/jquery.init.js',
+        'openwisp-notifications/js/object-notifications.js',
+    ]
+    css = {'all': ['openwisp-notifications/css/object-notifications.css']}
+
+
+def add_object_notification_widget():
```
nemesisdesign, line 15: this function is a bit magical. I would flag it as private (rename it to `_add_object_notification_widget`) so we can change it without worrying too much about backward compatibility in the future if we need it

openwisp_notifications/widgets.py
```diff
@@ -0,0 +1,39 @@
...
+from openwisp_notifications import settings as app_settings
+
+
+class IgnoreObjectNotificationWidgetMedia:
+    extend = True
+    js = [
+        'admin/js/jquery.init.js',
+        'openwisp-notifications/js/object-notifications.js',
+    ]
+    css = {'all': ['openwisp-notifications/css/object-notifications.css']}
```
nemesisdesign, line 12: could you convert these two lists to tuples for consistency with the other modules please?

openwisp_notifications/tasks.py
```diff
@@ -202,3 +215,11 @@ def ns_organization_created(instance_id):
     NotificationSetting.objects.bulk_create(
         notification_settings, ignore_conflicts=True
     )
+
+
+@shared_task
+def delete_ignore_object_notification_object(instance_id):
```
nemesisdesign, line 221: i'd rename it to `delete_ignore_object_notification`

openwisp_notifications/templates/admin/change_form.html
```diff
@@ -0,0 +1,24 @@
+{% extends 'admin/change_form.html' %}
+
+{% block admin_change_form_document_ready %}
+    {{ block.super }}
+    <script>
+        const owNotifyAppLabel = "{{ opts.app_label }}",
+            owNotifyModelName = "{{ opts.model_name }}",
```
nemesisdesign, line 7: here you're using double quotes but single quotes below, let's keep consistency and use single quotes

openwisp_notifications/templates/admin/change_form.html
```diff
@@ -0,0 +1,24 @@
+{% extends 'admin/change_form.html' %}
+
+{% block admin_change_form_document_ready %}
+    {{ block.super }}
```
nemesisdesign, line 4: I think we need to add a comment (using the django template comment syntax `{# comment #} that explains what this template does)

nemesisdesign (changes requested):
I'm not sure what went wrong, but while testing this branch on the latest openwisp-monitoring and openwisp-controller, the button disappeared :sob: 

Can you test the same and let me know?

pandafy:
I removed the changes from openwisp-monitoring which brings this feature as per your suggestion to move it to openwisp-controller altogether. For testing, please add following in `settings.py`
```
OPENWISP_NOTIFICATIONS_OBJECT_NOTIFICATION_WIDGET = ['openwisp_controller.config.admin.Device']
```

I will configure this in openwisp-controller in https://github.com/openwisp/openwisp-controller/pull/265 (I am currently working on it.)

README.rst
```diff
@@ -759,6 +769,26 @@ It sets the number of seconds the notification contents should be stored in the
 If you want cached notification content to never expire, then set it to ``None``.
 Set it to ``0`` if you don't want to store notification contents in cache at all.
 
+``OPENWISP_NOTIFICATIONS_OBJECT_NOTIFICATION_WIDGET``
```
nemesisdesign, line 772: The term "object notification" may not be super clear for a user, I'm thinking if we can make this more explicit... what about:

- `OPENWISP_NOTIFICATIONS_IGNORABLE_ADMIN_CLASSES`
- `OPENWISP_NOTIFICATIONS_IGNORABLE_OBJECTS`
- `OPENWISP_NOTIFICATIONS_IGNORE_ENABLED_ADMIN`

Or anything else along these lines? Any better idea?

openwisp_notifications/checks.py
```diff
@@ -27,3 +27,32 @@ def check_cors_configuration(app_configs, **kwargs):
...
+            checks.Warning(
+                msg='Improperly Configured',
+                hint=(
+                    'OW_OBJECT_NOTIFICATION_WIDGET should be a list,'
+                    f' {type(app_settings.OW_OBJECT_NOTIFICATION_WIDGET)} provided'
+                ),
+                obj='Settings',
+            )
+        )
+    for model_admin in app_settings.OW_OBJECT_NOTIFICATION_WIDGET:
```
nemesisdesign, line 46: model_admin is technically not an instance of `ModelAdmin` but a python `path`, while reading this code I thought it was checking if the path was an instance of `ModelAdmin` and got confused.

I think it would make the code more readable to rename `model_admin` to `path`.

openwisp_notifications/checks.py
```diff
@@ -27,3 +27,32 @@ def check_cors_configuration(app_configs, **kwargs):
...
+            errors.append(
+                checks.Error(
+                    msg='Improperly Configured',
+                    hint=(
+                        'OW_OBJECT_NOTIFICATION_WIDGET should contain dotted path string'
+                        f'to ModelAdmin, found {type(app_settings.OW_OBJECT_NOTIFICATION_WIDGET)}'
+                    ),
+                    obj='Settings',
+                )
+            )
```
nemesisdesign, line 57: then here we could import the class using [import_string](https://docs.djangoproject.com/en/3.1/ref/utils/#django.utils.module_loading.import_string) and ensure is an instance of `ModelAdmin` as well as mentioned above.

nemesisdesign (commented):
> I removed the changes from openwisp-monitoring which brings this feature as per your suggestion to move it to openwisp-controller altogether. For testing, please add following in `settings.py`
> 
> ```
> OPENWISP_NOTIFICATIONS_OBJECT_NOTIFICATION_WIDGET = ['openwisp_controller.config.admin.Device']
> ```

I tried setting it as per your suggestion but the button did not appear, then, after several minutes of debugging I realized it should be
`DeviceAdmin` instead of `Device`, which in turn made me think, why don't we perform a check on this?

We could raise a system check error if an element in the list is not an instance of django `ModelAdmin`, and advise a fix, which would make it easier for users to understand, what do you think?

The rest looks good!

pandafy:
> I tried setting it as per your suggestion but the button did not appear, then, after several minutes of debugging I realized it should be
> `DeviceAdmin` instead of `Device`, which in turn made me think, why don't we perform a check on this?

Apologies for misleading you. :sweat_smile: I have added the required checks.

@nemesisdesign did you notice that I had referred this feature as object notification in README when describing it to users. I tried referencing as "ignore object notification widget" but in my opinion it sounds a bit odd. Do you think this inconsistency is beneficial or will it only come haunting us in future?

README.rst
pandafy: `OPENWISP_NOTIFICATIONS_IGNORE_ENABLED_ADMIN` sounds good. :smile:

README.rst
```diff
@@ -586,6 +586,16 @@ by disabling both web and email option for a notification setting.
...
 respectively.
 
+Object Notification Preferences
+-------------------------------
+*OpenWISP Notifications* allows users to even define notification preferences based on target object.
+Users can disable notifications for any object for a desired period of time.
+Using the widget on an object's admin form, a user can disable notifications for that object
+for a day, week, month or even forever.
+
+**Note**: This feature requires configuring `"OPENWISP_NOTIFICATIONS_IGNORE_ENABLED_ADMIN" <#openwisp-notifications-object-notification-widget>`_
```
nemesisdesign, line 596: the internal reference seems mistaken, should be `openwisp_notifications_ignore_enabled_admin`

nemesisdesign (approved):
Thanks!

I found some details about the README to improve, will send improvements directly to dev shortly.

PR #124, opened 2020-09-07
[fix] Fixed accessibility issues of notification widget #122

### LINKED ISSUE #112: [ux] JS error when clicking on x icon of toast notification
Not sure exactly how I triggered it, but here's what I see in the console:

```
Uncaught TypeError: Cannot read property 'classList' of null
    at markNotificationRead (notifications.js:253)
    at HTMLDivElement.<anonymous> (notifications.js:312)
    at HTMLDocument.dispatch (jquery.js:5429)
    at HTMLDocument.elemData.handle (jquery.js:5233)
```

Lower priority.

### LINKED ISSUE #122: [ux] Clickable elements in notification widget should be <a> elements
Clickable elements in notification widget should be `<a>` elements for reasons of accessibility.

Each notification is a `<p>` which contains an `<a>`, the `<p>` could become `<a>` and the `<a>` could become something neutral like <span>.

pandafy:
Fixed notification bell and notification element not accessible from keyboard.
Added handlers for key events.

Closes #122

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,8 +12,9 @@ notifications:
   on_success: never
 
 env:
-  - DJANGO="django>=2.2,<3.0"
-  - DJANGO="django>=3.0,<3.1"
+  - DJANGO="django~=2.2"
+  - DJANGO="django~=3.0"
+  - DJANGO="django~=3.1"
 
 addons:
   apt:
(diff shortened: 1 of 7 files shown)


pandafy:
Hnet-image (14)

nemesisdesign (changes requested):
This is great!

I have one request, can you make the add the :focus style to so that it's the same as the :hover for all the clickable elements in the widget? This will cause the elements to change color as if the mouse is moving over them, which basically allows to browse the widget with the keyboard and have the same experience as with the mouse.

nemesisdesign (approved):
Thanks! :+1:

PR #140, opened 2020-10-19
[fix] Fixed playing sound from multiple windows #137

### LINKED ISSUE #137: [bug] Avoid playing sound if user is connected from more than one browser window/tab
When multiple tabs/browser windows are open and a notification is received, the sound is played multiple times.

I wonder if it's possible to find out how to avoid the multiple sound, but still ensure the notification is sent to each browser tab/window.

niteshsinha17:
In this solution, I am making use of localStorage in such a way that I am creating different IDs for different tabs and windows of a browser and storing it in the "windows" variable in localStorage. Whenever a notification comes I give access to play the sound to the window/tab whose ID comes first in the "windows" variable. Also removing its ID from the variable on page reload or close.

Fixes #137

### DIFF
--- a/openwisp_notifications/static/openwisp-notifications/js/notifications.js
+++ b/openwisp_notifications/static/openwisp-notifications/js/notifications.js
@@ -12,29 +12,43 @@ if (typeof gettext === 'undefined') {
         notificationWidget($);
         initNotificationDropDown($);
         initWebSockets($);
-        owNotificationWindow.set();
+        owNotificationWindow.init($);
     });
-    $(window).on('beforeunload', function () {owNotificationWindow.remove();});
-    $(window).on('storage', function () {owNotificationWindow.set();});
 })(django.jQuery);
 
 const owNotificationWindow = {
-    set: function(){
-        // Sets the first opened window
-        // Sets another window when first opened window get closed
-        if (localStorage.getItem('owWindow') === null) {
-            localStorage.setItem('owWindow',owWindowId);
-        }
+    // Following functions are used to decide which window has authority
+    // to play notification alert sound when multiple windows are open.
+
+    init: function init($) {
+        // Get authority to play notification sound
+        // when current window is in focus
+        $(window).on('focus', function() {owNotificationWindow.set();});
+
+        // Give up the authority to play sound before
+        // closing the window
+        $(window).on('beforeunload', function() {owNotificationWindow.remove();});
+
+        // Get authority to play notification sound when
+        // other windows are closed
+        $(window).on('storage', function () {
+            if (localStorage.getItem('owWindowId') === null) {
+                owNotificationWindow.set();
+            }
+        });
+    },
+    set: function() {
+        localStorage.setItem('owWindowId', owWindowId);
     },
     remove: function () {
-        // Removes the window then tab/window gets closed
-        if (localStorage.getItem('owWindow') === owWindowId){
-            localStorage.removeItem('owWindow');
+        if (localStorage.getItem('owWindowId') === owWindowId) {
+            localStorage.removeItem('owWindowId');
         }
     },
-    canPlaySound: function(){
-        // Returns true if the owWindow is the current window
-        return localStorage.getItem('owWindow') === owWindowId;
+    canPlaySound: function() {
+        // Returns whether current window has the authority to play
+        // notification sound
+        return localStorage.getItem('owWindowId') === owWindowId;
     },
 };
 
@@ -332,7 +346,7 @@ function initWebSockets($) {
                            </div>`);
             $('.ow-notification-toast-wrapper').prepend(toast);
             if (owNotificationWindow.canPlaySound()){
-                // To prevent playing of sound from multiple window
+                // Play notification sound only from authorized window
                 notificationSound.currentTime = 0;
                 notificationSound.play();
             }
--- a/openwisp_notifications/static/openwisp-notifications/js/notifications.js
+++ b/openwisp_notifications/static/openwisp-notifications/js/notifications.js
@@ -1,6 +1,7 @@
 'use strict';
 const notificationReadStatus = new Map();
 const userLanguage = navigator.language || navigator.userLanguage;
+const owWindowId = String(Date.now());
 
 if (typeof gettext === 'undefined') {
     var gettext = function(word){ return word; };
@@ -11,8 +12,31 @@ if (typeof gettext === 'undefined') {
         notificationWidget($);
         initNotificationDropDown($);
         initWebSockets($);
+        owNotificationWindow.set();
     });
+    $(window).on('beforeunload', function () {owNotificationWindow.remove();});
+    $(window).on('storage', function () {owNotificationWindow.set();});
 })(django.jQuery);
+
+const owNotificationWindow = {
+    set: function(){
+        // Sets the first opened window
+        // Sets another window when first opened window get closed
+        if (localStorage.getItem('owWindow') === null) {
+            localStorage.setItem('owWindow',owWindowId);
+        }
+    },
+    remove: function () {
+        // Removes the window then tab/window gets closed
+        if (localStorage.getItem('owWindow') === owWindowId){
+            localStorage.removeItem('owWindow');
+        }
+    },
+    canPlaySound: function(){
+        // Returns true if the owWindow is the current window
+        return localStorage.getItem('owWindow') === owWindowId;
+    },
+};
 
 function initNotificationDropDown($) {
     $('.ow-notifications').click(function (e) {
@@ -307,8 +331,11 @@ function initWebSockets($) {
                                 </div>
                            </div>`);
             $('.ow-notification-toast-wrapper').prepend(toast);
-            notificationSound.currentTime = 0;
-            notificationSound.play();
+            if (owNotificationWindow.canPlaySound()){
+                // To prevent playing of sound from multiple window
+                notificationSound.currentTime = 0;
+                notificationSound.play();
+            }
             toast.slideDown('slow', function () {
                 setTimeout(function () {
                     toast.slideUp('slow', function () {
--- a/openwisp_notifications/static/openwisp-notifications/js/notifications.js
+++ b/openwisp_notifications/static/openwisp-notifications/js/notifications.js
@@ -19,16 +19,13 @@ if (typeof gettext === 'undefined') {
 const owNotificationWindow = {
     // Following functions are used to decide which window has authority
     // to play notification alert sound when multiple windows are open.
-
     init: function init($) {
         // Get authority to play notification sound
         // when current window is in focus
-        $(window).on('focus', function() {owNotificationWindow.set();});
-
+        $(window).on('focus load', function() {owNotificationWindow.set();});
         // Give up the authority to play sound before
         // closing the window
         $(window).on('beforeunload', function() {owNotificationWindow.remove();});
-
         // Get authority to play notification sound when
         // other windows are closed
         $(window).on('storage', function () {


openwisp_notifications/static/openwisp-notifications/js/notifications.js
```diff
@@ -11,9 +12,56 @@ if (typeof gettext === 'undefined') {
...
+    let tempWindowsList = getWindowList();
+    let access_to = tempWindowsList[0];
+    if (access_to === windowId){
+        return true;
+    }
+    else {
+        return false;
+    }
+}
+
```
nemesisdesign, line 64: these functions could be rewritten in a more compact form:

```javascript
const notificationWindow = {
    add: function(){
        // code
    },
    remove: function(){
        // code
    },
    list: function(){
        // code
    },
    canPlaySound: function(){
        // code
    }
}
```

Less pollution of the global variable namespace.

openwisp_notifications/static/openwisp-notifications/js/notifications.js
```diff
@@ -11,9 +12,56 @@ if (typeof gettext === 'undefined') {
...
+function canPlaySound() {
+    // give access to play sound to first window of the list
+    let tempWindowsList = getWindowList();
+    let access_to = tempWindowsList[0];
+    if (access_to === windowId){
+        return true;
+    }
+    else {
+        return false;
+    }
```
nemesisdesign, line 62: simplify, just `return access_to === windowId`

openwisp_notifications/static/openwisp-notifications/js/notifications.js
```diff
@@ -11,9 +12,56 @@ if (typeof gettext === 'undefined') {
...
+    localStorage.windows = tempWindowsList.join(",");
+}
+
+function getWindowList() {
+    // get 'windows' list in from of array
+    var tempWindowsList = [];
+    if (localStorage.windows){
+        tempWindowsList = localStorage.windows.split(",");
+    }
+    return tempWindowsList;
```
nemesisdesign, line 50: simplify:

```javascript
if (localStorage.windows){
    return localStorage.windows.split(",");
}
return [];
```

openwisp_notifications/static/openwisp-notifications/js/notifications.js
```diff
@@ -11,9 +12,56 @@ if (typeof gettext === 'undefined') {
...
+
+function addToWindowList() {
+    // add window id to 'windows' in localStorage
+    windowId = String(Date.now());
+    let tempWindowsList = getWindowList();
+    let index = tempWindowsList.indexOf(windowId);
+    if (index < 0){
+        tempWindowsList.push(windowId);
+    }
+    localStorage.windows = tempWindowsList.join(',');
```
nemesisdesign, line 31: please use `localStorage.notificationWindows`, which is more specific and easy to debug

nemesisdesign (changes requested):
Nice solution, I tested it and it seems to work, @pandafy what do you think?

@niteshsinha17 the code needs improvements, see my comments below.

openwisp_notifications/static/openwisp-notifications/js/notifications.js
```diff
@@ -11,9 +12,56 @@ if (typeof gettext === 'undefined') {
         notificationWidget($);
         initNotificationDropDown($);
         initWebSockets($);
+        addToWindowList();
     });
 })(django.jQuery);
 
+window.addEventListener("beforeunload", function () {
```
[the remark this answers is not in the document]
niteshsinha17: Okay @nemesisdesign , I will make the changes as you suggested.

openwisp_notifications/static/openwisp-notifications/js/notifications.js
niteshsinha17: Hi @nemesisdesign , I saw your comments after doing my last amend.  I also like the way you suggested and I did the same. But it is giving error during tests again. In my last amend I had declared `canPlaySound` in `addEventListener` which is in `initWebSockets` function and then tests were passing. What do you think about it?

nemesisdesign: what error is it giving?

niteshsinha17: I tested it again now it is not giving any error. I have made changes as you have suggested. Please review it once again.

nemesisdesign (commented):
@niteshsinha17 I tested it again and applied some improvements, please can you double check?

While testing it, I noticed that in some rare cases, it may happen that a window is not removed correctly. Can we look for a solution to deal with this?
We could loop over the windows, parse the date and remove any entry that is older than something like 12 hours. How does that sound?

niteshsinha17:
@nemesisdesign I will review it once again. 

Your solution is good but I have one more. What if we give access to play sound to the window which is at the last? It will always have the most recent window opened.

nemesisdesign:
> @nemesisdesign I will review it once again.
> 
> Your solution is good but I have one more. What if we give access to play sound to the window which is at the last? It will always have the most recent window opened.

That sounds great, thank you!
We should still try to remove old dates to avoid filling the localStorage space in case anything goes wrong with the removal, what don't you think?

niteshsinha17:
Okay, we can add a function `clearOldWindows`  which we run before adding a window and it will remove all old windows that are older than 12 hours.

nemesisdesign:
> Okay, we can add a function `clearOldWindows` which we run before adding a window and it will remove all old windows that are older than 12 hours.

That sounds great!

openwisp_notifications/static/openwisp-notifications/js/notifications.js
```diff
@@ -11,9 +12,61 @@ if (typeof gettext === 'undefined') {
...
+        // remove windows older than 12 hours
+        let tempWindowsList = notificationWindow.list();
+        let newWindowsList = tempWindowsList.map((window)=>{
+            if(((+windowId)-(+window))/36e5<=12){
+                return window;
+            }
+        });
+        localStorage.notificationWindows = newWindowsList.join(',');        
+    }
+};
```
pandafy, line 66:
```suggestion
const oWNotificationWindow = {
    set: function(){
        // Sets the last focused window
        localStorage.setItem('owLastWindow', owWindowId);
    },
    remove: function () {
        // Removes the last focused window entry if current window
        // was referenced
        if (localStorage.getItem('owLastWindow') === owWindowId){
            localStorage.removeItem('owLastWindow');
        }
    },
    canPlaySound: function(){
        // Returns true if the current window was the last fouces one
        return localStorage.getItem('owLastWindow') === owWindowId;
    },
};
```

I tried tried this implementation where we never store more than one `windowId` in `localStorage` at any time. This implementation has following benefits:

- It is quicker. We are not dealing with array, we are reducing time and space complexity. 
- It does not bloats `localStorage` as concerned by @nemesisdesign. At any time, there will be atmost one instance of the last focused window. Even if the value is not removed while closing a tab, it will get overridden when a new window is opened or focus in changed to another window.

I didn't understand to keep track of all opened windows in the first place. I think the problem is allow sound to be played only from one window only, We need not worry about how many windows were opened by the user. @niteshsinha17 do you still think we need to maintain the list of opened windows?

openwisp_notifications/static/openwisp-notifications/js/notifications.js
```diff
@@ -307,8 +360,10 @@ function initWebSockets($) {
                                 </div>
                            </div>`);
             $('.ow-notification-toast-wrapper').prepend(toast);
-            notificationSound.currentTime = 0;
-            notificationSound.play();
+            if (notificationWindow.canPlaySound()){
```
pandafy, line 363: It will be better to add a comment here mention the reason why this condition is being checked here.

openwisp_notifications/static/openwisp-notifications/js/notifications.js
```diff
@@ -11,9 +12,61 @@ if (typeof gettext === 'undefined') {
         notificationWidget($);
         initNotificationDropDown($);
         initWebSockets($);
+        notificationWindow.clearOldWindows();
+        notificationWindow.add();
     });
+
```
pandafy, line 18:
```suggestion

       $(window).on('focus', function () {oWNotificationWindow.set();});
```

This will return the authority of notifications to be played to the last window which had focus instead of the last opened window. With the current implementation I observed that the notification sound was getting played in the different window even when I had currently focused on another window.

pandafy (changes requested):
@niteshsinha17 some suggestion to treasure for future contributions as well

- Never use generic identifiers(e.g. `windowId`, `notificationWindow`) in the global  namespace. This will help avoid any conflicts which can arise in the future when someone uses those same identifiers while extending OpenWISP and utilises them for completely different purpose. 
- Use `const` for storing data whose value is not supposed to be changes during the execution of program.

I have suggested a modified implementation for fixing the bug which is built on your patch. Feel free to ask questions if you have any trouble with them.

openwisp_notifications/static/openwisp-notifications/js/notifications.js
niteshsinha17: The purpose of storing old windows was the following reasons:

- To change the authority of playing sound if needed.

- Suppose if all opened windows of openwisp-notifications are not focused and the user suddenly closes the last window during being on some other website then the authority should be given to 2nd last or another window to insure that sound plays when a notification is received. It's not possible if we don't store all windows id. 

- Suppose two windows of the same browser are opened not on tab different tabs but on two different windows then it is possible to focus on both windows at the same time by minimizing the window size. Then which window will play sound in this situation. 

User can do anything, I have tried to target all the possibilities that a user can do. The Idea you suggested was also in my mind but because of these problems, I thought to store window id. It definitely increases time and space complexity but the size of the array is not very large.

pandafy: > Suppose if all opened windows of openwisp-notifications are not focused and the user suddenly closes the last window during being on some other website then the authority should be given to 2nd last or another window to insure that sound plays when a notification is received. It's not possible if we don't store all windows id.

Good observation. In that case, we can just add another event listener to handle it. Whenever there is a change in localStorage, `storage.onChange` event is fired. We can handle this event to assign the id of the current window to the localStorage. I acknowledge this will create a race condition, but since in this case, we can play the sound from any window randomly that should not be a concern.

```

   $(window).on('storage', function () {
        if (localStorage.getItem('owLastWindow') === null) {
            oWNotificationWindow.set();
        }
    });
```

> Suppose two windows of the same browser are opened not on tab different tabs but on two different windows then it is possible to focus on both windows at the same time by minimizing the window size. Then which window will play sound in this situation.

Did you test this? I tested this scenario, but the sound was getting played from only one window. Focus is something different from which window is currently occupying the screen. Can you confirm whether you are still able to reproduce it? If you do, share the steps you took to reproduce this. 

> User can do anything, I have tried to target all the possibilities that a user can do. 

I agree, we should try to cover all the cases. But, we can give up on few rare ones like "what if the notification arrives during the execution of storage event handler" to avoid bloating the code.

nemesisdesign: @pandafy from the suggested change it's not immediately evident to me what happens when the user closes the window which plays the sound? How is another windows chosen to play the sound?

niteshsinha17: > Good observation. In that case, we can just add another event listener to handle it. Whenever there is a change in localStorage, storage.onChange event is fired. We can handle this event to assign the id of the current window to the localStorage. I acknowledge this will create a race condition, but since in this case, we can play the sound from any window randomly that should not be a concern.

Here you are calling `set` inside the change handler, It will start an infinite loop which will change `window id` infinite times. This doesn't look good.

niteshsinha17: > Did you test this? I tested this scenario, but the sound was getting played from only one window. Focus is something different from which window is currently occupying the screen. Can you confirm whether you are still able to reproduce it? If you do, share the steps you took to reproduce this.

Sorry, I hadn't tested it. I was thinking focus is something related to visibility and I was wrong.

niteshsinha17: I have one more solution which doesn't have any race condition or doesn't store all window id.
In this, I am storing window Id as 0,1,2,3... Code will become as follows:
```
const owNotificationWindow = {
    set: function(){
        // set window id to 'notificationWindows' in localStorage
        // and returns the window id
        localStorage.setItem('owWindow',+localStorage.getItem('owWindow')+1);
        return +localStorage.getItem('owWindow');
    },
    remove: function(){
        // remove window id from 'notificationWindows'
        // in localStorage on reload and close
        localStorage.setItem('owWindow',+localStorage.getItem('owWindow')-1);
    },
    canPlaySound: function(){
        // returns true if the current window is
        // same as saved in localStorage
        return +localStorage.getItem('owWindow') === owWindowId;
    },
    changed: function(){
        // change window id if window id in localStorage
        // is greater than this window id
        if(owWindowId > +localStorage.getItem('owWindow')){
            owWindowId-=1;
        }
    }
};
var owWindowId = owNotificationWindow.set();
```
```
(function ($) {
    $(document).ready(function () {
        notificationWidget($);
        initNotificationDropDown($);
        initWebSockets($);
    });

    $(window).on('beforeunload', function () {owNotificationWindow.remove();});
    $(window).on('storage', function () {
        owNotificationWindow.changed();
    });
})(django.jQuery);
```

```
if (owNotificationWindow.canPlaySound()){
                // prevent playing of sound from multiple window
                notificationSound.currentTime = 0;
                notificationSound.play();
            }
```
In this solution, we need to use var in the global variable. Please help to find the best one according to you.

pandafy: > Here you are calling set inside the change handler, It will start an infinite loop which will change window id infinite times. This doesn't look good.

@niteshsinha17  again, did you test it? If this is the case, share the steps you took to reproduce it.

I could not find the link earlier but here is the [W3 specification](https://www.w3.org/TR/2008/WD-html5-20080610/structured.html) for firing `Storage.onChange` event

> When the setItem(), removeItem(), and clear() methods are called on a Storage object x that is associated with a local storage area, then in every HTMLDocument object whose Window object's localStorage attribute's Storage object is associated with the same storage area, other than x, a storage event must be fired, as described below.

Look for it in section 4.11.1.4. 

> @pandafy from the suggested change it's not immediately evident to me what happens when the user closes the window which plays the sound? How is another windows chosen to play the sound?

@nemesisdesign the new event handler will manage this case

> Suppose if all opened windows of openwisp-notifications are not focused and the user suddenly closes the last window during being on some other website then the authority should be given to 2nd last or another window to insure that sound plays when a notification is received. It's not possible if we don't store all windows id.

When the window is closed in this scenario, the `Storage.onChange` event will be fired in other windows. The authority of playing sound will be given to either the first one to complete the execution or the last one to set key in localStorage.

niteshsinha17: > @niteshsinha17 again, did you test it? If this is the case, share the steps you took to reproduce it.
@pandafy  again sorry, I not noticed it `localStorage.getItem('owLastWindow') === null`. I will not do it again. According to me solution you have given is the best. Can I do changes as you have suggested?

niteshsinha17: > @niteshsinha17 again, did you test it? If this is the case, share the steps you took to reproduce it.


@pandafy again sorry, I not noticed it localStorage.getItem('owLastWindow') === null. I will not do it again. According to me solution you have given is the best. Can I do changes as you have suggested?

niteshsinha17: > @niteshsinha17 again, did you test it? If this is the case, share the steps you took to reproduce it.

@pandafy again sorry, I haven't noticed it `localStorage.getItem('owLastWindow') === null`. Its not good. I will try to avoid it next time. 

I like your solution. Can I do changes as you suggested?

nemesisdesign (commented):
@niteshsinha17 what's the status of this one?

niteshsinha17:
> @niteshsinha17 what's the status of this one?

I was waiting for your reply. I think the solution given by @pandafy is best. I am going to do the changes now.

niteshsinha17:
I have changed it a little bit. Storing the first window instead of last please review it

openwisp_notifications/static/openwisp-notifications/js/notifications.js
```diff
@@ -11,9 +12,32 @@ if (typeof gettext === 'undefined') {
...
     });
+    $(window).on('beforeunload', function () {owNotificationWindow.remove();});
+    $(window).on('storage', function () {owNotificationWindow.set();});
 })(django.jQuery);
 
+const owNotificationWindow = {
+    set: function(){
+        // Sets the first opened window
+        // Sets another window when first opened window get closed
+        if (localStorage.getItem('owWindow') === null) {
```
pandafy, line 25: Doing this check here will break the implementation. 

When a user will change focus to another window, value of `localStorage.getItem('owWindow')` will not be `null`, hence the the notification alert won't be played from the right window. 

Why did you make this change?

pandafy (commented):
I am currently looking into why the build is falling.

openwisp_notifications/static/openwisp-notifications/js/notifications.js
niteshsinha17: I though playing of sound is more important. Also when in previous code when last opened window gets closed then also sound is not played from the focused window. 

If you think last opened window should play sound then we can change it. But user can traverse to any window at any time that's why I didn't consider it too change every time when a new window is opened.

nemesisdesign (commented):
I merged master in this PR to fix the build and it seems it's working (still some time for some jobs to finish).

Last time I had tested this it was working, is it ready for final review/testing?

Guys, if it solves the problem and there's no major issues let's just merge it, I'd like to resolve issue #137 and include this fix in the next minor release of openwisp-notifications, which will be released via ansible-openwisp2 to our users.

nemesisdesign (commented):
While testing it I detected a bug.

Open the admin page, create a notification, then start navigating around the admin and create a notification after changing page, the sound doesn't play at all.

If I type `owWindowId` in the JS console and then create a notification, the sound comes back.

Adding "onload" along with "focus" fixes it.

nemesisdesign (approved):
Merging, if you see any issue with my addition of the load event please send another patch.

Thanks @niteshsinha17 and @pandafy! :+1:

PR #143, opened 2020-11-05
[fix] Fixed stalling of migrate command when celery broker is unreachable #142

### LINKED ISSUE #142: [migrations] Migrations wait indefinitely when redis is not running
@pandafy informed that there are `post_migrate` hook when migrating database.
If redis is not running the system would not report any error but silently wait for the redis to come online.

This is bad design because the developer may not realize that the migrations is waiting for redis.

We should:
- inform the developer that redis is required for running migrations and fail if redis is not running.
- update docs so that redis setup is before the running migrations step.

pandafy:
Solution:

- Decreased max_retry_limit for celery tasks
- Added error handling for the error thrown by celery for broker being unreachable
- Added logger to notify errror to the user and provide information on resolving the error
    
Note:

- An active bug in celery does not let the developer to selectively set max_retry_limit for a particular task.
  See https://github.com/celery/celery/issues/4627
- In this patch, a workaround is implemented which sets the global max_rety_limit. This decreases max_retry_limit for every task.
     
Fixes #142

### DIFF
--- a/tests/openwisp2/settings.py
+++ b/tests/openwisp2/settings.py
@@ -162,6 +162,11 @@ else:
     CELERY_TASK_EAGER_PROPAGATES = True
     CELERY_BROKER_URL = 'memory://'
 
+# Workaround for stalled migrate command
+CELERY_BROKER_TRANSPORT_OPTIONS = {
+    'max_retries': 10,
+}
+
 CELERY_BEAT_SCHEDULE = {
     'delete_old_notifications': {
         'task': 'openwisp_notifications.tasks.delete_old_notifications',
--- a/README.rst
+++ b/README.rst
@@ -946,6 +946,41 @@ Delete object notification setting
 
     DELETE /api/v1/notification/ignore/{app_label}/{model_name}/{object_id}/
 
+
+Management Commands
+-------------------
+
+``populate_notification_preferences``
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+This command will populate notification preferences for all users for organizations
+they are member of.
+
+Example usage:
+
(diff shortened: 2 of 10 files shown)


nemesisdesign (commented):
I forgot to submit this review! Not sure if it's still relevant, just submitting with delay, let me know if any point has been already addressed.

tests/openwisp2/settings.py
```diff
@@ -162,6 +162,11 @@
     CELERY_TASK_EAGER_PROPAGATES = True
     CELERY_BROKER_URL = 'memory://'
 
+# Workaround for stalled migrate command
+CELERY_BROKER_TRANSPORT_OPTIONS = {
+    'max_retries': 10,
```
atb00ker, line 167: Patch, looks good, my only question is how much time would it take for the `10` retries?
Or in other words, how long would the user wait before seeing the error?

atb00ker (commented):
Great going, just one question below:

nemesisdesign (commented):
```

**Review by nemesisdesign — CHANGES_REQUESTED** [review]
Looks good! The build is failing for a flake8 warning:

```./openwisp_notifications/tests/test_notifications.py:12:1: F401 'django.test.override_settings' imported but unused
```

Please fix it so we can merge.

tests/openwisp2/settings.py
pandafy: For 10 tries, it will take around 1.5 minutes. If it sounds a lot, we can just lower this number. I forgot to mention that we will have to add this in the test project of openwisp-controller as well.

nemesisdesign (approved):
@pandafy go ahead and merge

Issue #145, opened 2020-11-06
[CI] Build failing because of celery 3


nemesisdesign:
Copy/pasting what I wrote i n the [general chat](https://gitter.im/openwisp/general).

```
Collecting channels<4
  Downloading channels-3.0.1-py3-none-any.whl (38 kB)
```

celery 3.0.x is being installed, which was released 7 days ago and is not compatible.
Something is causing this to be installed. Maybe the django-redis version you pinned is not correct.
I thought the build was failing, I must have merged by mistake. Anyway, now you should know where to look to fix this.

PR #147, opened 2020-11-06
[chores] Added Github Action for CI

pandafy:
- Removed Travis integration


runtests.py
```diff
@@ -20,4 +20,4 @@
         args.insert(2, 'openwisp2')
         app_dir = ['tests', 'openwisp2', 'sample_notifications']
     execute_from_command_line(args)
-    sys.exit(pytest.main([os.path.join(*app_dir, 'tests', 'test_websockets.py')]))
+    # sys.exit(pytest.main([os.path.join(*app_dir, 'tests', 'test_websockets.py')]))
```
atb00ker, line 23: This is commented out. :smile:

pandafy: This is has been done deliberately due to https://github.com/openwisp/openwisp-notifications/issues/145. The build will stall if I don't comment out this code which will make it hard for us to judge this.

.github/workflows/build.yml
```diff
@@ -0,0 +1,88 @@
...
+      with:
+        ref: ${{ github.event.pull_request.head.sha }}
+
+    - name: Set up Python ${{ matrix.python-version }}
+      uses: actions/setup-python@v2
+      with:
+        python-version: ${{ matrix.python-version }}
+
+    - name: Install system packages
+      run: sudo apt -y install sqlite3 gdal-bin
```
dwang, line 47:
```suggestion
      run: sudo apt-get -y install sqlite3 gdal-bin
```

[`WARNING: apt does not have a stable CLI interface. Use with caution in scripts.`](https://github.com/openwisp/openwisp-notifications/runs/1364607740?check_suite_focus=true#step:5:8)

We can make it less noisy by using the `-qq` flag if you want. Also, do you think we need to run `apt-get update` before as well? It works right now but I'm not sure if something might break in the future.

.github/workflows/build.yml
```diff
@@ -0,0 +1,88 @@
...
+
+    - name: Install python dependencies dependencies
```
dwang, line 49:
```suggestion
    - name: Install python dependencies
```

runtests.py
nemesisdesign: Now that #145 is merged, can this line be uncommented?

nemesisdesign (approved):
First impression is great! :+1: 

See my comment below, I think we can proceed after that.

PR #151, opened 2020-11-13
[fix] Fixed compatible versioning in dependencies and tests for websockets

pandafy:
Bug:
Tests for websockers were importing ASGI application router object from
test project files. This implementation will break if tests are in
another project.

Fix:
Imported the ASGI application router usin string from settings object

Refs https://github.com/openwisp/openwisp-controller/pull/316#discussion_r517749794

### DIFF
--- a/openwisp_notifications/tests/test_websockets.py
+++ b/openwisp_notifications/tests/test_websockets.py
@@ -4,9 +4,10 @@ from unittest.mock import patch
 import pytest
 from channels.db import database_sync_to_async
 from channels.testing import WebsocketCommunicator
+from django.conf import settings
 from django.contrib.auth import get_user_model
 from django.contrib.contenttypes.models import ContentType
-from openwisp2.routing import application
+from django.utils.module_loading import import_string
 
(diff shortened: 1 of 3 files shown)


nemesisdesign (commented):
Looks good!

The build is inexplicably failing because openwisp-utils 0.7, I am wondering how is it ending being installed if it's pinned down in the requirements?

pandafy:
The build is failing because of my broken understanding of `compatible versioning`. `openwisp-users~=0.6` is equivalent to `openwisp-users==0.*`. I will fix this in a PR for upgrading dependencies.

pandafy:
>  I will fix this in a PR for upgrading dependencies.

A slight change of plans. I fixed it in this PR. We won't be able to upgrade to openwisp-utils 0.7.0 until openwisp-users 0.5 is released. That meant we would be stalling this PR for no reason. 

I hope the dependencies are pinned properly this time.

nemesisdesign (approved):
Go ahead :+1:

PR #158, opened 2020-12-26
[feature] Added general setting to disable all notifications #148

### LINKED ISSUE #148: [feature] Allow to disable notifications for all organizations or keep everything disabled except notifications for specific organizations
Current buggy situation:

**Possible solution:**

niteshsinha17:
Created a new model named `General Setting` through which we can disable email OR/AND web notifications for all the organization whether it is created after or before disabling.

Also, I have written tests for it. 
I think we may need to write code for API too and also need to mention this feature in the documentation. 
Before doing this it will be great if you give consent to this solution.   

closes #148


pandafy:
@niteshsinha17 the changes that you have made in this PR were not expected. 

Read this discussion to get a better idea of what is expected from the issue: https://github.com/openwisp/openwisp-notifications/pull/69#pullrequestreview-471188099

niteshsinha17:
> @niteshsinha17 the changes that you have made in this PR were not expected.
> 
> Read this discussion to get a better idea of what is expected from the issue: [#69 (review)](https://github.com/openwisp/openwisp-notifications/pull/69#pullrequestreview-471188099)

Okay, I will check and will make changes according to it.

niteshsinha17:
@pandafy from this https://github.com/openwisp/openwisp-notifications/pull/69#pullrequestreview-471188099 I have concluded a few things that I wanted to share with you.

-  I need to create that `All` setting which was removed in this PR but this time the need is to disable all notifications?
- If yes then we will again come into the confusing situation which was mentioned in the above PR.
>  if notifications are enabled in ALL but not enabled in the org, which setting overrides which?

If the above points are true then I think this could be a solution.
- That 'All' setting should not generate automatically. It will be created by the user by creating a new notification setting to disable all notifications.
- If a user disables notifications for all organizations for a particular type then we can check the record before setting the notification.
- If he enables notifications both web and email for all organizations then we will delete that setting.

What do you think about it?

pandafy:
@nemesisdesign will be the right person to answer this query.

nemesifier:
Closing as we're working on this in GSoC24. Thanks Nitesh!

PR #167, opened 2021-01-19
[bug] Fix IntegrityError in celery task(ns_user_created) when staff user is created

R9295:


nemesisdesign (changes requested):
Checking the logs again, I see that the task which is causing the failure is `ns_organization_user_added_or_updated` (not `ns_user_created` which you edited here).

nemesisdesign (changes requested):
@R9295 the change does not fix the issue, this change does:

```
@receiver(post_save, sender=User, dispatch_uid='user_notification_setting')
def notification_setting_user_created(instance, created, **kwargs):
    transaction.on_commit(
        lambda: tasks.ns_user_created.delay(
            instance.pk,
            instance.is_superuser,
            created
        )
    )
```

This is the receiver which is failing.

I suggest changing only this handler and see what happens with the tests.

I believe sooner or later we may have to apply the same change to most receivers, because on some operations we have to ensure the background tasks are executed on up to date data, which can be ensured only by using `transaction.on_commit`, which causes the task to be scheduled only once the SQL query has been committed to the DB.

@pandafy I would like you to be aware of this issue we are facing since it may require changes to many handlers, for now maybe we'll patch only one but it's very probable that other similar bugs may show up for the other handlers too. I caught this one because it's the most obvious one, but it's hard to spot because it's happening only in prod.

PR #168, opened 2021-01-24
[feature] Introduced mechanism to clear cache on specific signals #164

### LINKED ISSUE #164: [feature] Introduce mechanism to clear the cache on specific signals
When items are renamed, old name still shows up.

We should introduce a generic mechanism to listen for change signals of objects that are used in notifications, so that the cache can be re-generated for all elements related to that specific object.

Question: it is possible to clear the cache only for specific objects or would we need to clear the entire cache?

We could allow other modules to register their own model and signals to listen to, eg:

```
from openwisp_controller.config.signals import device_name_changed 

register_notification_observation((Device, device_name_changed))
```

This could go in ow-controller and would trigger a refresh of the cache when the device name is changed.

I think this is similar to https://github.com/openwisp/openwisp-notifications/issues/163, maybe we can introduce a solution which can handle both cases, we need a solution to allow other modules to tell ow-notifications when to refresh the cache in an efficient way.

PS: the `device_name_changed` signal does not exist yet, but we can add it once the ideas in this issue are confirmed, see https://github.com/openwisp/openwisp-controller/issues/361.

niteshsinha17:
closes #164


openwisp_notifications/handlers.py
```diff
@@ -316,3 +317,19 @@ def schedule_object_notification_deletion(instance, created, **kwargs):
...
             (instance.pk,), eta=instance.valid_till
         )
+
+
+def register_notification_observation(model, signal):
+    signal.connect(
+        update_notification_cache,
+        sender=model,
+        dispatch_uid='{}_{}_update_notification_cache'.format(
+            str(signal), model.__name__
```
devkapilbansal, line 327: Can't we use `f-strings` here?

nemesisdesign: I agree.

tests/openwisp2/sample_notifications/apps.py
```diff
@@ -22,3 +26,9 @@ def register_notification_types(self):
                 'email_subject': '[{site.name}] INFO: {notification.target} created',
             },
         )
+
+    def connect_recievers(self):
+        from openwisp_notifications.handlers import register_notification_observation
```
nemesisdesign, line 31: is it necessary to import this here?

openwisp_notifications/handlers.py
nemesisdesign: another thing, are we sure `str(signal)` returns the name of the signal? Eg: `post_save`. Please double check this.

openwisp_notifications/handlers.py
```diff
@@ -316,3 +317,19 @@ def schedule_object_notification_deletion(instance, created, **kwargs):
         tasks.delete_ignore_object_notification.apply_async(
             (instance.pk,), eta=instance.valid_till
         )
+
+
+def register_notification_observation(model, signal):
```
nemesisdesign, line 322: I'd call this `register_notification_cache_update` or anything similar which makes it evident that by passing a model and a signal, the notification cache will be updated when that signal from that model is emitted.

nemesisdesign (changes requested):
@niteshsinha17 good work here, I left some comments below, please also add a basic test and add a brief doc section in the README which explains what problem this function solves, the section can be added as a subsection of [Notification Cache](https://github.com/openwisp/openwisp-notifications#notification-cache).

openwisp_notifications/handlers.py
niteshsinha17: > Can't we use f-strings here?

Yes we can use it. But in most of the places `.format()` is being used. 

> another thing, are we sure str(signal) returns the name of the signal? Eg: post_save. Please double check this.

No, It do not return signal name. I have added another parameter named `signal_uid` for this.

niteshsinha17:
@nemesisdesign  I have added test and document in it. Please check. Thank you

openwisp_notifications/handlers.py
devkapilbansal: > Yes we can use it. But in most of the places `.format()` is being used.
>

Then we should open an issue for this. :thinking:

niteshsinha17: I don't think so. It is present in many modules. I have seen it. Then we will need to open an issue in each module. I thinks it's not harming code in anyway. To do this we must have a strong reason for it.

openwisp_notifications/tests/test_notifications.py
```diff
@@ -822,3 +829,17 @@ def test_post_migrate_handler_celery_broker_unreachable(self, mocked_logger, *ar
...
+        self._register_notification_cache_update(User, post_save, 'post_save')
+        self.notification_options.update(
+            {'action_object': operator, 'target': operator, 'type': 'default'}
+        )
+        self._create_notification()
+        content_type = ContentType.objects.get_for_model(operator._meta.model)
+        operator.username = "new operation name"
+        operator.save()
+        cache_key = Notification._cache_key(content_type.id, operator.id)
+        _cache = cache.get(cache_key, None)
```
nemesisdesign, line 844: there is this assertion which checks that the cache has been emptied but there's no assertion which checks the cache contains anything, or am I wrong? We should have an assertion that ensures the cache contains the right data

README.rst
```diff
@@ -385,6 +385,32 @@ The cached values are updated automatically to reflect actual data from database
...
+
+    def ready(self):
+        super().ready()
+
+        # Include lines after this inside
+        # ready function of you app config class
+        from openwisp_notifications.handlers import register_notification_cache_update
+
+        model = load_model('app_name', 'model_name')
+        register_notification_cache_update(model, post_save, signal_uid="post_save")
```
nemesisdesign, line 407: ensure quoting style is consistent (single quotes instead of double quotes, or the opposite, look at the rest of the examples in the docs to decide).

Please also change `signal_uid="post_save"` to have a more unique name, otherwise users may be copy/paste this code which is potentially problematic since post_save is too generic to be a uid.

Eg: 

```python
register_notification_cache_update(
    model,
    post_save,
    signal_uid='myapp_mymodel_notification_cache_invalidation'
)
```

README.rst
```diff
@@ -385,6 +385,32 @@ The cached values are updated automatically to reflect actual data from database
 the duration of caching these objects using
 `OPENWISP_NOTIFICATIONS_CACHE_TIMEOUT setting <#OPENWISP_NOTIFICATIONS_CACHE_TIMEOUT>`_.
 
+**register_notification_cache_update**
+This function is used to register a signal of a model which is being used as an ``actor``, ``action_object``
+and ``target`` objects. As these values are cached for the optimization purpose so their cached values are
```
nemesisdesign, line 390: Let's change the beginning of this section to something like:

``` 
Cache invalidation
~~~~~~~~~~~~~~~~~~

The function ``register_notification_cache_update`` can be used to register...
```

nemesisdesign (changes requested):
Good progress @niteshsinha17, see my comments below.

tests/openwisp2/sample_notifications/apps.py
niteshsinha17: Sorry I forget to answer this last time. 
Yes, it is necessary because when I was importing it at the top of the file it was giving error that apps not loaded.

openwisp_notifications/tests/test_notifications.py
niteshsinha17: Yes, you are right. Actually I was thinking deletion of cache is important in this test. 
I will add test to ensures that the cache contains the right data

README.rst
niteshsinha17: Okay I will change them. Thanks

niteshsinha17:
@nemesisdesign I have made changes.

nemesisdesign (commented):
Looks good @niteshsinha17!

I think we should proceed with the implementation of https://github.com/openwisp/openwisp-controller/issues/361 (pointing the build to use this branch) before merging this, so we can be sure this implementation really solves our issue.

I want to remind you what the issue is: right now the issue is that when the name of a device is changed, the notifications keep referring to it with the old name.

When a new device registers, it registers using the mac address as the name, then an operator usually changes its name to something more human readable, but the notification keep referring to it with the mac address.

pandafy (changes requested):
I know I am very late to the party(or probably early for the next one 😆). 

I came here from reviewing https://github.com/openwisp/openwisp-controller/pull/407

I think we should add one more test that uses a custom signal. This will check the purpose this feature was added for.

openwisp_notifications/tests/test_notifications.py
```diff
@@ -822,3 +829,20 @@ def test_post_migrate_handler_celery_broker_unreachable(self, mocked_logger, *ar
...
+        self._create_notification()
+        content_type = ContentType.objects.get_for_model(operator._meta.model)
+        operator.username = 'new operator name'
+        operator.save()
+        cache_key = Notification._cache_key(content_type.id, operator.id)
+        _cache = cache.get(cache_key, None)
+        self.assertEqual(_cache, None)
+        self._create_notification()
+        new_cache = cache.get(cache_key, None)
+        self.assertEqual(new_cache.username, 'new operator name')
```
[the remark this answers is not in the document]
nemesisdesign: Thanks @pandafy, I think you're right, I missed these details, @niteshsinha17 please follow Gagan's recommendations when you can :blush:

openwisp_notifications/handlers.py
```diff
@@ -316,3 +317,19 @@ def schedule_object_notification_deletion(instance, created, **kwargs):
...
+        dispatch_uid='{}_{}_update_notification_cache'.format(
+            signal_uid, model.__name__
+        ),
+    )
+
+
+def update_notification_cache(sender, instance, **kwargs):
+    content_type = ContentType.objects.get_for_model(instance)
+    cache_key = Notification._cache_key(content_type.id, instance.id)
+    cache.delete(cache_key)
```
nemesisdesign, line 335: here's another thing I didn't spot before, we should call these operations after previous changes have been committed to the database using `transaction.on_commit`, otherwise we risk having different behaviour in prod than what we have in development.

In production, there are many occasions in which sigals fire before changes have been committed to the DB, leading to bugs. We fixed most of these issues but we're still fixing others: https://github.com/openwisp/openwisp-notifications/pull/167

Therefore it's better to prevent this here.

Could be done like the following:

```suggestion
from django.db import transaction

def update_notification_cache(sender, instance, **kwargs):
    def invalidate_cache():
        content_type = ContentType.objects.get_for_model(instance)
        cache_key = Notification._cache_key(content_type.id, instance.id)
        cache.delete(cache_key)
    # execute cache invalidation only after changes have been committed to the DB
    transaction.on_commit(invalidate_cache)
```

nemesisdesign (changes requested):

openwisp_notifications/handlers.py
niteshsinha17: @nemesisdesign Thanks

README.rst
```diff
@@ -385,6 +385,34 @@ The cached values are updated automatically to reflect actual data from database
...
+        # ready function of you app config class
+        from openwisp_notifications.handlers import register_notification_cache_update
+
+        model = load_model('app_name', 'model_name')
+        register_notification_cache_update(model, post_save, signal_uid="myapp_mymodel_notification_cache_invalidation")
+
+**Note**: You need to import ``register_notification_cache_update`` inside the ``ready`` function or 
+you can define another funtion to register signals which will be called in ``ready`` and then it will be 
+imported in this function. Also ``signal_uid`` is unique identifier of a signal. You can pass any 
+value you want but it is suggested to pass signal name as a string. By default it is ``None``.
```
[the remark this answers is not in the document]
niteshsinha17: > I think we should add one more test that uses a custom signal. This will check the purpose this feature was added for.

@pandafy Should I use custom signal instead of ``post_save`` or I have to do something else?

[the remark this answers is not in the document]
pandafy: Use a custom signal for a test in sample app.

niteshsinha17:
@pandafy Its I have made changes according to you. 
I have also created a new file named ``test_transaction_notifications.py`` because of [this fact](https://docs.djangoproject.com/en/3.1/topics/db/transactions/#use-in-tests). Some tests were failing because of this so I have created a separate file of transaction tests.

openwisp_notifications/tests/test_transaction_notifications.py
```diff
@@ -0,0 +1,51 @@
...
+            verb='Test Notification',
+            email_subject='Test Email subject',
+            url='https://localhost:8000/admin',
+        )
+
+    def _create_notification(self):
+        return notify.send(**self.notification_options)
+
+    def _register_notification_cache_update(self, model, signal, dispatch_uid):
+        register_notification_cache_update(model, signal, dispatch_uid=dispatch_uid)
```
pandafy, line 32: I don't see the use of this method. It is simply calling the other function. Why is this required?

openwisp_notifications/tests/test_transaction_notifications.py
```diff
@@ -0,0 +1,51 @@
...
+    def test_notification_cache_update(self):
+        operator = self._get_operator()
+        self._register_notification_cache_update(
+            User, post_save, 'operator_name_changed_invalidation'
+        )
+        self.notification_options.update(
+            {'action_object': operator, 'target': operator, 'type': 'default'}
+        )
+        self._create_notification()
+        content_type = ContentType.objects.get_for_model(operator._meta.model)
```
pandafy, line 43:
```suggestion
        content_type = ContentType.objects.get_for_model(User)
```

openwisp_notifications/tests/test_transaction_notifications.py
```diff
@@ -0,0 +1,51 @@
...
+        cache_key = Notification._cache_key(content_type.id, operator.id)
+        operator_cache = cache.get(cache_key, None)
+        self.assertEqual(operator_cache.username, operator.username)
+        operator.username = 'new operator name'
+        operator.save()
+        notification = Notification.objects.get(target_content_type=content_type)
+        operator_cache = cache.get(cache_key, None)
+        self.assertEqual(notification.target.username, 'new operator name')
```
pandafy, line 51:
```suggestion
        notification = Notification.objects.get(target_content_type=content_type)
        operator_cache = cache.get(cache_key, None)
        self.assertEqual(notification.target.username, 'new operator name')  # Done for populating cache
        self.assertEqual(operator_cache.username, 'new operator name')
```

openwisp_notifications/tests/test_transaction_notifications.py
```diff
@@ -0,0 +1,51 @@
...
+from openwisp_notifications.handlers import register_notification_cache_update
+from openwisp_notifications.signals import notify
+from openwisp_notifications.swapper import load_model
+from openwisp_users.tests.utils import TestOrganizationMixin
+
+User = get_user_model()
+Notification = load_model('Notification')
+
+
+class TestTransactionNotifications(TestOrganizationMixin, TransactionTestCase):
```
pandafy, line 16: We don't need to create another file for this. You can move this class to `test_notifications.py`. I believe it will make more sense there.

tests/openwisp2/sample_notifications/tests/test_django.py
```diff
@@ -46,6 +53,30 @@ def test_test_app_object_created_notification(self):
...
+    def test_test_app_cache_invalidation(self):
+        test_app = self._get_test_app()
+        content_type = ContentType.objects.get_for_model(test_app._meta.model)
+        cache_key = Notification._cache_key(content_type.id, test_app.id)
+        test_app_cache = cache.get(cache_key, None)
+        self.assertEqual(test_app_cache.name, test_app.name)
+        test_app.name = 'new test app'
+        test_app.save()
+        notification = Notification.objects.get(target_content_type=content_type)
+        self.assertEqual(notification.target.name, test_app.name)
```
pandafy, line 77:
```suggestion
        self.assertEqual(notification.target.name, test_app.name)
        test_app_cache = cache.get(cache_key, None)
        self.assertEqual(test_app_cache.name, test_app.name)
```

You need to test the value in cache as well.

README.rst
```diff
@@ -406,12 +406,12 @@ To register a signal you need to include following code in your ``apps.py``.
...
         model = load_model('app_name', 'model_name')
-        register_notification_cache_update(model, post_save, signal_uid="myapp_mymodel_notification_cache_invalidation")
+        register_notification_cache_update(model, post_save, dispatch_uid="myapp_mymodel_notification_cache_invalidation")
 
 **Note**: You need to import ``register_notification_cache_update`` inside the ``ready`` function or 
 you can define another funtion to register signals which will be called in ``ready`` and then it will be 
-imported in this function. Also ``signal_uid`` is unique identifier of a signal. You can pass any 
-value you want but it is suggested to pass signal name as a string. By default it is ``None``.
+imported in this function. Also ``dispatch_uid`` is unique identifier of a signal. You can pass any 
+value you want but it needs to be unique. For more details read `preventing-duplicate-signals <https://docs.djangoproject.com/en/dev/topics/signals/#preventing-duplicate-signals>`_
```
pandafy, line 414: I think it is better that we explicitly tell users that they will be going to an external website by clicking on this link. 

```suggestion
value you want but it needs to be unique. For more details read `preventing duplicate signals section of Django documentation<https://docs.djangoproject.com/en/dev/topics/signals/#preventing-duplicate-signals>`_
```

pandafy (changes requested):
Good progress @niteshsinha17! This PR is almost complete, requested some changes.

openwisp_notifications/tests/test_notifications.py
```diff
@@ -822,3 +826,42 @@ def test_post_migrate_handler_celery_broker_unreachable(self, mocked_logger, *ar
...
+        self._create_notification()
+        content_type = ContentType.objects.get_for_model(operator._meta.model)
+        cache_key = Notification._cache_key(content_type.id, operator.id)
+        operator_cache = cache.get(cache_key, None)
+        self.assertEqual(operator_cache.username, operator.username)
+        operator.username = 'new operator name'
+        operator.save()
+        notification = Notification.objects.get(target_content_type=content_type)
+        cache_key = Notification._cache_key(content_type.id, operator.id)
+        self._create_notification()
```
niteshsinha17, line 863: @pandafy without this I was getting this error
```
AttributeError: 'NoneType' object has no attribute 'username'
```

nemesisdesign (approved):
Great work! :+1:

PR #170, opened 2021-03-04
[chores] Minor fixes to HTML

nemesisdesign:
- removed type="text/javascript" which in HTML5 is not needed
- added required alt attribute to bell icon


pandafy (approved):
@nemesisdesign I have fixed the failing tests and the changes looks good.

PR #175, opened 2021-03-15
[fix] Avoid multiple jquery inclusions in base_site.html

nemesisdesign:
Avoid including jquery if block.super already includes it

### DIFF
--- a/openwisp_notifications/templates/admin/base_site.html
+++ b/openwisp_notifications/templates/admin/base_site.html
@@ -9,7 +9,7 @@
 
 {% block extrahead %}
   {{ block.super }}
-  {% if not media.js %}
+  {% if not media.js and 'jquery' not in block.super %}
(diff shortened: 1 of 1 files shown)


nemesisdesign (commented):
Not sure what's wrong with the build.

@pandafy we need this change because I found out the dashboard in openwisp-utils doesn't load (because jQuery is not loaded) if the notifications module is not loaded, therefore depending on which module is loaded first in INSTALLED_APPS, jquery may be included two times without this fix, the same is also added to openwisp-utils: https://github.com/openwisp/openwisp-utils/pull/171/files#diff-14471639a21bc28b7dce88ac39528df7c5165048bfb59c9d0ede785c96dc944aR14-R18.

pandafy (approved):
LGTM 👍🏼 

@nemesisdesign merge it at your discretion. Or, we can merge it along with https://github.com/openwisp/openwisp-utils/pull/171 and test it more.

PR #177, opened 2021-04-14
[ux] Fix notification icon

pniaps:
This commit fixes the bell icon after changes from https://github.com/openwisp/openwisp-utils/pull/182

Before:

https://user-images.githubusercontent.com/7120526/114720743-934da200-9d38-11eb-9c49-21b4709eb5a1.mp4

After:

https://user-images.githubusercontent.com/7120526/114720765-96e12900-9d38-11eb-8a8a-08be811ce762.mp4

### DIFF
--- a/openwisp_notifications/static/openwisp-notifications/css/notifications.css
+++ b/openwisp_notifications/static/openwisp-notifications/css/notifications.css
@@ -295,16 +295,6 @@
   }
 }
 
-@media (max-width: 1024px) {
-  .ow-notifications {
-    display: block !important;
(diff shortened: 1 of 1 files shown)


pniaps:
I guess it was added because in admin_theme, the user icon was going to the bottom left as seen in https://github.com/openwisp/openwisp-utils/pull/182#issue-614693480, so this code moved it top right, next to the menu button.

This change should not break anything since the only element that has the class "ow-notifications" is that one.

nemesisdesign (approved):

PR #180, opened 2021-05-12
[bug] Error handling send_email_notification #179

### LINKED ISSUE #179: [bug] Add error handling in send_email_notification handler
If a notification setting does not exist in the database, the following code will raise `DoesNotExist` exception. 
https://github.com/openwisp/openwisp-notifications/blob/a20957b086fa10ab49a4488a231c3e1c926b617a/openwisp_notifications/handlers.py#L173-L177

**Possible solution**

If such an exception is raised, return from the function i.e. don't send any email

**Notes for writing failing tests**

1. Ensure the recipient of the notification does not have any NotificationSetting
2. Instead of using `self._create_notification` helper of the `TestNotifications` class create a notification using `Notification.objects.create()`. The former will not create a notification if NotificationSetting does not exist in the database

RahulPalve:
Added exception handling in send_email_notification, So if
`notification_setting = instance.recipient.notificationsetting_set.get(
                organization=target_org, type=instance.type
            )` 
NotificationSetting Does Not Exist it will return from function and mail will not be sent.

Closes #179

### DIFF
--- a/openwisp_notifications/handlers.py
+++ b/openwisp_notifications/handlers.py
@@ -171,9 +171,12 @@ def send_email_notification(sender, instance, created, **kwargs):
     # Get email preference of user for this type of notification.
     target_org = getattr(getattr(instance, 'target', None), 'organization_id', None)
     if instance.type and target_org:
-        notification_setting = instance.recipient.notificationsetting_set.get(
-            organization=target_org, type=instance.type
-        )
+        try:
+            notification_setting = instance.recipient.notificationsetting_set.get(
+                organization=target_org, type=instance.type
+            )
+        except NotificationSetting.DoesNotExist:
+            return
         email_preference = notification_setting.email_notification
     else:
         # We can not check email preference if notification type is absent,


nemesisdesign (changes requested):
Great, but a test for the bug is missing.

RahulPalve:
hey @nemesisdesign, I didn't understand  "test case for a bug". Should I write a test case which is checking that code is generating `DoesNotExist` Exception in case  `notification_setting` does not exist? I am new to openwisp-notification please assist.

nemesisdesign (commented):
> hey @nemesisdesign, I didn't understand "test case for a bug". Should I write a test case which is checking that code is generating `DoesNotExist` Exception in case `notification_setting` does not exist? I am new to openwisp-notification please assist.

Hi @RahulPalve,

whenever we fix bug, the first thing we do is to write a test that fails and replicate the bug. You can write it also afterwards, you just need to comment your changes and restore the old version, write the failing test, then uncomment the fix and the test should pass.

Is it clearer now?

pandafy (changes requested):
Tests are missing. If you have any trouble in writing tests, feel free to ask for help in [OpenWISP's Gitter chat](https://gitter.im/openwisp/general).

pandafy:
Closing this due to lack of activity.

PR #192, opened 2021-09-17
[fix] Fixed creating and updating OrganizationUser deleted existing NotificationSettings #190

### LINKED ISSUE #190: [bug] Notification preferences are lost
It happened a few times to me after upgrading some instances that some notification preferences were lost.
Since this has happened a few times I think there is a bug but I am not exactly sure where it is.

pandafy:
Closes #190


openwisp_notifications/handlers.py
```diff
@@ -274,16 +274,58 @@ def notification_type_registered_unregistered_handler(sender, **kwargs):
...
+            org_id=instance.organization_id,
+        )
+
+
+@receiver(
+    pre_save,
+    sender=OrganizationUser,
+    dispatch_uid='update_orguser_notification_setting',
+)
+def notification_setting_org_user_org_updated(instance, **kwargs):
```
pandafy, line 295: This is very rudimentary code which have further scope to be improved(optimised). I wrote it to realise the bug and to explicitly share the corner cases which were missing in the previous implementation. 

I have also tried to take a different approach that does not require these much computation or code. I opened https://github.com/pandafy/openwisp-notifications/pull/4 for proof of concept.

openwisp_notifications/handlers.py
```diff
@@ -274,16 +274,58 @@ def notification_type_registered_unregistered_handler(sender, **kwargs):
...
     dispatch_uid='create_orguser_notification_setting',
 )
-def notification_setting_org_user_created(instance, **kwargs):
-    if instance.is_admin:
-        tasks.ns_organization_user_added_or_updated.delay(
-            instance_id=instance.pk,
-            instance_user_id=instance.user_id,
-            instance_org_id=instance.organization_id,
+def notification_setting_org_user_created(instance, created, **kwargs):
+    if created and instance.is_admin:
```
nemesisdesign, line 282: what if the OrganizationUser is changed and is_admin is set from False to True? Then notification settings wouldn't be created right?
Is that solved with the new receiver below?

openwisp_notifications/handlers.py
```diff
@@ -274,16 +274,58 @@ def notification_type_registered_unregistered_handler(sender, **kwargs):
...
+        )
+
+
+@receiver(
+    pre_save,
+    sender=OrganizationUser,
+    dispatch_uid='update_orguser_notification_setting',
+)
+def notification_setting_org_user_org_updated(instance, **kwargs):
+    # This method does not handles a case where user of
```
nemesisdesign, line 296: why is this not a celery task?

openwisp_notifications/handlers.py
pandafy: Yes, the `pre_save` receiver takes care of that.

openwisp_notifications/base/admin.py
```diff
@@ -15,14 +15,28 @@ def get_readonly_fields(self, request, obj=None):
...
+
     def has_add_permission(self, request, obj=None):
         return request.user.is_superuser
 
     def has_delete_permission(self, request, obj=None):
-        return request.user.is_superuser
+        # A notification is considered deleted when the delete
+        # boolean field is set. A user can never delete
+        # NotificationSettting from the database.
+        return False
```
nemesisdesign, line 31: are you sure this won't prevent us from deleting users? Did you try to delete users? I remember in the past we had problems when setting this to False (we couldn't delete the related objects from the admin).

openwisp_notifications/base/models.py
```diff
@@ -203,6 +204,7 @@ class AbstractNotificationSetting(UUIDModel):
     email = models.BooleanField(
         _('email notifications'), null=True, blank=True, help_text=_(_RECEIVE_HELP)
     )
+    deleted = models.BooleanField(_('Delete'), null=True, blank=True, default=False,)
```
nemesisdesign, line 207: there's a final comma which should be removed here

openwisp_notifications/tasks.py
```diff
@@ -204,17 +182,9 @@ def ns_organization_created(instance_id):
...
     for user in User.objects.filter(is_superuser=True):
-        for type in notification_types:
-            notification_settings.append(
-                NotificationSetting(
-                    user_id=user.id, type=type, organization_id=instance_id
-                )
-            )
-    NotificationSetting.objects.bulk_create(
-        notification_settings, ignore_conflicts=True
-    )
```
pandafy, line 217: I think this code can be kept, since there will be no existing notification setting for this organization. Hence, bulk_create will be better here.

pandafy:
With the latest changes, the notification preferences will works as following 

- A user can effectively delete notification preferences from the database. These notification preferences will be re-created by the automation. There is a [dedicated section mentioning this in the README](https://github.com/openwisp/openwisp-notifications/blob/69c7d304a52302aa9e5e303b0f878889cf7ed8ca/README.rst#deleting-notification-preferences).
- Automation will never remove notification preference from the database. It will flag notification preference as "deleted"
- Notification preferences flagged as "deleted" will not appear in user admin interface 
- Notifications will be not be send to user if the preference is marked as deleted.

README.rst
```diff
@@ -610,6 +610,13 @@ by disabling both web and email option for a notification setting.
 then ``email_notification`` or ``web_notification`` option of concerned notification type will be used
 respectively.
 
+Deleting Notification Preferences
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Deleting the notification preferences is an advanced option. Users should turn off web and email
+notifications instead of deleting notification preferences. Deleted notification preferences
+might get re-created by automation.
```
nemesisdesign, line 618: what about "may be re-created automatically if the system needs it."? Or something along this line.

openwisp_notifications/tasks.py
nemesisdesign: Let's keep the new version, so we don't have to maintain different versions of the same code.

nemesisdesign (commented):
Looks good, I will test it in the next days.

nemesisdesign (changes requested):
I tried monkey patching `ns_register_unregister_notification_type` but it cannot be done easily: importing it in the settings raises django.core.exceptions.AppRegistryNotReady: Apps aren't loaded yet.

This means that even with this patch, the issue we are having will keep happening (it will be only easier to fix).

Please add a way to disable the receiver which is triggered on post_migrate with a setting (in a separate commit).

nemesisdesign (commented):
Looks good, let's test it a bit more before merging.

nemesisdesign (approved):

PR #198, opened 2021-10-04
[change] Added "models" parameter to register_notification_type #163

### LINKED ISSUE #163: [bug] Too many delete_obsolete_objects tasks
In production systems, I observe too many `delete_obsolete_objects` being executed, which is unnecessary.

We should find a way to limit these.

One possibility would be to allow registering models which are sensitive to the pre_delete signals and allow each module to add their models, so we'd have:

- User model added by default by openwisp-notifications
- Device model added by default by openwisp-controller

These at the moment are the only two modules we need to listen to pre-delete events for.

pandafy:
- delete_obsolete_objects task will excuted for models that are
  mentioned in the models list

Closes #163

### DIFF
--- a/openwisp_notifications/handlers.py
+++ b/openwisp_notifications/handlers.py
@@ -9,7 +9,7 @@ from django.core.mail import EmailMultiAlternatives
 from django.db import transaction
 from django.db.models import Q
 from django.db.models.query import QuerySet
-from django.db.models.signals import post_delete, post_save, pre_delete
+from django.db.models.signals import post_delete, post_save
 from django.dispatch import receiver
 from django.template.loader import render_to_string
 from django.utils import timezone
@@ -19,7 +19,10 @@ from openwisp_notifications import settings as app_settings
 from openwisp_notifications import tasks
 from openwisp_notifications.exceptions import NotificationRenderException
 from openwisp_notifications.swapper import load_model, swapper_load_model
-from openwisp_notifications.types import get_notification_configuration
+from openwisp_notifications.types import (
+    NOTIFICATION_ASSOCIATED_MODELS,
+    get_notification_configuration,
+)
 from openwisp_notifications.websockets import handlers as ws_handlers
 
 logger = logging.getLogger(__name__)
@@ -232,12 +235,14 @@ def clear_notification_cache(sender, instance, **kwargs):
     )
 
 
-@receiver(pre_delete, dispatch_uid='delete_obsolete_objects')
+@receiver(post_delete, dispatch_uid='delete_obsolete_objects')
 def related_object_deleted(sender, instance, **kwargs):
     """
     Delete Notification and IgnoreObjectNotification objects having
     "instance" as related object.
     """
+    if sender not in NOTIFICATION_ASSOCIATED_MODELS:
+        return
     instance_id = getattr(instance, 'pk', None)
     if instance_id:
         instance_model = instance._meta.model_name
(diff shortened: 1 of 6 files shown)


openwisp_notifications/handlers.py
```diff
@@ -232,12 +235,14 @@ def clear_notification_cache(sender, instance, **kwargs):
...
 
-@receiver(pre_delete, dispatch_uid='delete_obsolete_objects')
+@receiver(post_delete, dispatch_uid='delete_obsolete_objects')
 def related_object_deleted(sender, instance, **kwargs):
     """
     Delete Notification and IgnoreObjectNotification objects having
     "instance" as related object.
     """
+    if sender not in NOTIFICATION_ASSOCIATED_MODELS:
+        return
```
pandafy, line 245: The [build is failing](https://github.com/openwisp/openwisp-notifications/runs/3793141179?check_suite_focus=true#step:11:69) due to following test: 

https://github.com/openwisp/openwisp-notifications/blob/015bd4685d883d01c5fb2640bc1beac83b9cef2e/openwisp_notifications/tests/test_ignore_object_notification.py#L44-L52

The above code wass also responsible for deleting `IgnoreObjectNotification` objects related to the deleted object. Now it is not working since we are executing the delete tasks only for the models that provided on registering notifications. 

**Thinking out loud**

The IgnoreNotificationSetting only makes sense for objects that will have a notification. Since such object/model will already have a notification type, this condition in above test will still be covered for most cases. 

I suggest opening another issue to implement registering of models along with [OPENWISP_NOTIFICATIONS_IGNORE_ENABLED_ADMIN](https://github.com/openwisp/openwisp-notifications#openwisp_notifications_ignore_enabled_admin) setting. Or, we can do that in below code without further developer intervention since we have path to the admin class .

https://github.com/openwisp/openwisp-notifications/blob/015bd4685d883d01c5fb2640bc1beac83b9cef2e/openwisp_notifications/widgets.py#L13-L46

pandafy: I have checked the code of ModelAdmin, and found that `Model` parameter is added when the ModelAdmin class is instantiated. The solution I proposed above will not be possible.

nemesisdesign: it doesn't sound like a big deal, if we need execute the task for some models anyway, let's just do it.

For example, this seems to work:

```python
if sender not in NOTIFICATION_ASSOCIATED_MODELS and sender not in [OrganizationUser, IgnoreObjectNotification]:
        return
```

Is `OrganizationUser` special from the point of view of the implementation or is it something specific to the test project app?

nemesisdesign (changes requested):
See my comment below.

Next step will be to update the code in OpenWISP Controller: https://github.com/openwisp/openwisp-controller/issues/557.

openwisp_notifications/handlers.py
pandafy: The [IgnoreObjectNotification model also has a generic foreign key field](https://github.com/openwisp/openwisp-notifications/blob/dc2d250f6f8b367a83aefbff27d6b65f563e45da/openwisp_notifications/base/models.py#L266-L267). It is very similar to the implementation of Notification model. 

As I shared in my previous comment: 

> The IgnoreNotificationSetting only makes sense for objects that will have a notification. Since such object/model will already have a notification type, this condition in above test will still be covered for most cases.

I will re-write the test to explain it better.

nemesisdesign: ok, now is clear, thanks

nemesisdesign (approved):

PR #199, opened 2021-10-05
[fix] IntegrityError in celery task when staff user is created # 165

### LINKED ISSUE #165: [bug] IntegrityError when staff user is created
I can replicate this in an instance installed via ansible-openwisp2.
I suspect it can be replicated also in the development environment with the celery worker running.

How to replicate:

- create a staff user, of the group administrator, add a relation to an org and flag (is_admin=True)

Expected outcome:

- notification settings are created

Actual outcome:

This error is logged:

```
IntegrityError: insert or update on table "openwisp_notifications_notificationsetting" violates foreign key constraint "openwisp_notificatio_user_id_3cc7bf95_fk_openwisp_"
DETAIL:  Key (user_id)=(13201df6-2bb2-4017-a95b-30a4e88a21a7) is not present in table "openwisp_users_user".
```

This error is really typical. It happens when the background task is passed to the worker before the object it must operate on is committed to the DB.

The solution is to use transaction.on_commit to call the celery task, like we do in other modules, eg:

- https://github.com/openwisp/openwisp-firmware-upgrader/blob/94288de5c970ab7fcd11a178d54bf7a45fcc98cc/openwisp_firmware_upgrader/base/models.py#L127-L129
- https://github.com/openwisp/openwisp-firmware-upgrader/blob/94288de5c970ab7fcd11a178d54bf7a45fcc98cc/openwisp_firmware_upgrader/base/models.py#L307

codesankalp:
The task should run after transaction commit.

Fixes #165


nemesisdesign (changes requested):
@codesankalp please see https://github.com/openwisp/openwisp-notifications/pull/167#pullrequestreview-587982314

codesankalp:
> @codesankalp please see [#167 (review)](https://github.com/openwisp/openwisp-notifications/pull/167#pullrequestreview-587982314)

I have seen that and `ns_user_created` is changed in https://github.com/openwisp/openwisp-notifications/commit/015bd4685d883d01c5fb2640bc1beac83b9cef2e# 

image

PR #208, opened 2021-10-31
[change] Switch to openwisp-utils email template #206

### LINKED ISSUE #206: [change] Switch to openwisp-utils email template
The email theme logic is being moved to openwisp-utils (https://github.com/openwisp/openwisp-utils/pull/263), to validate the work being done there we shall change the code of this module to use the utility function being moved there and ensure tests pass without changes.

devkapilbansal:
Closes #206


openwisp_notifications/tests/test_notifications.py
```diff
@@ -576,7 +578,7 @@ def test_notification_invalid_message_attribute(self, mocked_task):
         mocked_task.assert_called_with(notification_id=notification.id)
         unregister_notification_type('test_type')
 
-    @patch.object(app_settings, 'OPENWISP_NOTIFICATIONS_HTML_EMAIL', False)
+    @patch.object(utils_settings, 'OPENWISP_HTML_EMAIL', False)
```
devkapilbansal, line 581: These tests are already covered in openwisp-utils. I think we should remove them too :thinking:

nemesisdesign: definitely remove tests that have been moved to openwisp-utils.

openwisp_notifications/handlers.py
```diff
@@ -199,23 +196,14 @@ def send_email_notification(sender, instance, created, **kwargs):
         target_url = None
     if target_url:
         description += '\n\nFor more information see {0}.'.format(target_url)
```
nemesisdesign, line 198: let's take advantage to flag `For more information see` as translatable, which doesn't seem to flagged as translatable now.

Please also add "Find out more" since I asked you to remove it from openwisp-utils.

devkapilbansal: We are using `blocktrans` in our template which I think will take care of that.
Do we still need to do so here?

nemesisdesign: This one? https://github.com/openwisp/openwisp-utils/blob/issues/259-email-template/openwisp_utils/admin_theme/templates/openwisp_utils/email_template.html#L104
I didn't see that. It's totally useless, translations do not work that way, each string has to be flagged on its own.
Please study how translations work in Django and gettext here: https://docs.djangoproject.com/en/3.2/topics/i18n/translation/, read carefully the parts in which it shows how to deal with placeholder of text which is inserted in the strings and how that has to be written  in order to work (f stings and str.format() do not work).

openwisp_notifications/handlers.py
```diff
@@ -195,13 +196,16 @@ def send_email_notification(sender, instance, created, **kwargs):
     else:
         target_url = None
     if target_url:
-        description += '\n\nFor more information see {0}.'.format(target_url)
+        description += _('\n\nFor more information see {0}.'.format(target_url))
```
nemesisdesign, line 199: This won't work for the reasons expressed here: https://github.com/openwisp/openwisp-notifications/pull/208#discussion_r741145797

nemesisdesign (changes requested):

openwisp_notifications/handlers.py
devkapilbansal: >Python’s str.format() method will not work when either the format_string or any of the arguments to str.format() contains lazy translation objects. Instead, you can use django.utils.text.format_lazy(), which creates a lazy object that runs the str.format() method only when the result is included in a string. 

This is what the doc mentions. str.format doesn't work when there are lazy objects. But, I don't think we are translating `target_url` too. :thinking:

nemesisdesign: That's a parameter that has to be passed, pelase read the django docs.

openwisp_notifications/handlers.py
```diff
@@ -198,24 +196,20 @@ def send_email_notification(sender, instance, created, **kwargs):
...
     if target_url:
-        description += '\n\nFor more information see {0}.'.format(target_url)
-    mail = EmailMultiAlternatives(
-        subject=subject,
-        body=strip_tags(description),
-        from_email=settings.DEFAULT_FROM_EMAIL,
-        to=[instance.recipient.email],
+        description += _('\n\nFor more information see %(target_url)s.') % {
+            'target_url': target_url
+        }
```
nemesisdesign, line 201: This message was added only in text mode while here it seems it's being always added.
I think we need to change `send_email` to have an HTML body and a text body, so we can pass both.

nemesisdesign (changes requested):

nemesisdesign (commented):
From local testing looks good. I am waiting to test everything in a real system once the work in openwisp-users is ready.

nemesisdesign (approved):

PR #209, opened 2021-11-24
[fix] Fixed notification storm in UI #183

### LINKED ISSUE #183: [bug] Deal with notification storm
Related to #132
When an important event happens in the network (eg: blackout) and notifications are generated for hundred of devices, the following problems happen:

- the screen fills with toast notifications which seem to be infinite
- the sound keeps buzzing
- trying to open the notification widget fails because the UI is reloaded every second so it always shows loading

We should deal with these kind of events in a way that does not overwhelm users who have the UI open and are checking the situation, they know something big is happening and they do not need to be alerted for every notification. They also want to be able to check the notification widget properly (now it gets reloaded whenever a new notification comes up so the loading indicator shows up instead of notifications pretty much every second).

pandafy:
Closes #183

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -776,6 +776,31 @@ model, then configure the setting as following:
 This setting allows to disable creating `notification preferences <#notification-preferences>`_
 on running migrations.
 
+``OPENWISP_NOTIFICATIONS_NOTIFICATION_STORM_PREVENTION``
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+A ``dict`` containing configuration for notification storm prevention mechanism.
+The default configuration is as follows:
+
+.. code-block:: python
+
+    OPENWISP_NOTIFICATIONS_NOTIFICATION_STORM_PREVENTION = {
+        # Time period for tracking burst of notifications (in seconds)
+        'short_term_time_period': 10,
+        # Number of notifications considered as a notification burst
+        'short_term_notification_count': 6,
+        # Time period for tracking notifications in long time interval (in seconds)
+        'long_term_time_period': 60,
+        # Number of notifications in long time interval to be considered as a notification storm
+        'long_term_notification_count': 30,
+        # Initial time for which notification updates should be skipped (in seconds)
+        'initial_backoff': 1,
+        # Time by which skipping of notification updates should be increased (in seconds)
+        'backoff_increment': 1,
+        # Maximum interval after which the notification widget should get updated (in seconds)
+        'max_allowed_backoff': 15,
+    }
+
 Exceptions
 ----------
 
--- a/README.rst
+++ b/README.rst
@@ -779,7 +779,11 @@ on running migrations.
 ``OPENWISP_NOTIFICATIONS_NOTIFICATION_STORM_PREVENTION``
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-A ``dict`` containing configuration for notification storm prevention mechanism.
+When the system starts creating a lot of notifications because of a
+general network outage (e.g. : a power outage), the notification storm prevention
+mechanism avoids constantly displaying notifications.
+This setting allows tweaking the working of the mechanism.
+
 The default configuration is as follows:
 
 .. code-block:: python
--- a/README.rst
+++ b/README.rst
@@ -780,9 +780,14 @@ on running migrations.
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 When the system starts creating a lot of notifications because of a
-general network outage (e.g. : a power outage), the notification storm prevention
-mechanism avoids constantly displaying notifications.
-This setting allows tweaking the working of the mechanism.
+general network outage (e.g.: a power outage, a global misconfiguration),
+the notification storm prevention mechanism avoids the constant displaying
+of new notification alerts as well as their sound, only the notification
+counter will continue updating periodically, although it won't emit any
+sound or create any other visual element until the
+notification storm is over.
+
+This setting allows tweaking how this mechanism works.
 
 The default configuration is as follows:
 
--- a/README.rst
+++ b/README.rst
@@ -790,7 +790,7 @@ The default configuration is as follows:
         # Number of notifications considered as a notification burst
         'short_term_notification_count': 6,
         # Time period for tracking notifications in long time interval (in seconds)
-        'long_term_time_period': 60,
+        'long_term_time_period': 180,
         # Number of notifications in long time interval to be considered as a notification storm
         'long_term_notification_count': 30,
         # Initial time for which notification updates should be skipped (in seconds)
--- a/openwisp_notifications/settings.py
+++ b/openwisp_notifications/settings.py
@@ -20,6 +20,19 @@ IGNORE_ENABLED_ADMIN = getattr(
 POPULATE_PREFERENCES_ON_MIGRATE = getattr(
     settings, 'OPENWISP_NOTIFICATIONS_POPULATE_PREFERENCES_ON_MIGRATE', True
 )
+NOTIFICATION_STORM_PREVENTION = getattr(
+    settings,
+    'OPENWISP_NOTIFICATIONS_NOTIFICATION_STORM_PREVENTION',
+    {
+        'short_term_time_period': 10,
+        'short_term_notification_count': 6,
+        'long_term_time_period': 60,
+        'long_term_notification_count': 30,
+        'initial_backoff': 1,
+        'backoff_increment': 1,
+        'max_allowed_backoff': 15,
+    },
+)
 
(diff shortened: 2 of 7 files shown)


pandafy:
https://user-images.githubusercontent.com/32094433/143299615-8ca01dad-c8f6-47a3-b54b-701f5b3e2928.mp4

openwisp_notifications/websockets/handlers.py
```diff
@@ -11,12 +16,20 @@ def notification_update_handler(reload_widget=False, notification=None, recipien
         notification = NotificationListSerializer(notification).data
     except (NotFound, AssertionError):
         pass
+    is_notification_storm = reload_widget and (
+        Notification.objects.filter(
+            recipient=recipient,
+            timestamp__gte=now() - timedelta(10),
+        ).count()
+        > 10
+    )
```
pandafy, line 25: A notification storm is observed when more than 10 notifications are created for a user in last 10 seconds. (1 notification/second). 

I used this value from my personal experience of using OpenWISP. I have seen checks failing for a couple of devices at a time. So to avoid catching such occurrences as "notification storm", I used above value.

pandafy, line 25:
```suggestion
    is_notification_storm = reload_widget and (
        Notification.objects.filter(
            recipient=recipient,
            timestamp__gte=now() - timedelta(10),
        ).count()
        > 6
    )
```

pandafy:
I am using this script to test it

```python
import time

from django.core.management import call_command

for count in range(1, 101):
    call_command('create_notification')
    if (count % 20 == 0):
        time.sleep(20)
```

openwisp_notifications/websockets/consumers.py
```diff
@@ -37,20 +44,48 @@ def disconnect(self, close_code):
...
-            self.send(
-                json.dumps(
-                    {
-                        'type': 'notification',
-                        'notification_count': unread_notifications,
-                        'reload_widget': event['reload_widget'],
-                        'notification': event['notification'],
-                    }
+        if event['recipient'] != str(user.pk):
+            return
```
pandafy, line 48: This is the only line that has missing coverage and it is intentional. I have an idea in mind which will allow us to remove this condition from here. Hence, I didn't add a test for it. 
I will open a new issue for this with more details, once I test the idea.

openwisp_notifications/websockets/handlers.py
```diff
@@ -1,8 +1,28 @@
...
 
 from openwisp_notifications.api.serializers import NotFound, NotificationListSerializer
 
+from ..swapper import load_model
+
+Notification = load_model('Notification')
+
+
+def user_in_notification_storm(user):
+    in_notification_storm = cache.get(f'ow-noti-storm-{user.pk}', False)
```
nemesisdesign, line 14: I think we should add a docstring with a summary of how this works, don't you think?

openwisp_notifications/websockets/handlers.py
```diff
@@ -1,8 +1,28 @@
...
+    if in_notification_storm:
+        return True
+    qs = Notification.objects.filter(recipient=user)
+    if qs.filter(timestamp__gte=now() - timedelta(seconds=10)).count() >= 6:
+        in_notification_storm = True
+    elif qs.filter(timestamp__gte=now() - timedelta(seconds=60)).count() >= 30:
```
nemesisdesign, line 20: why 60 and not something longer? Like 300 seconds (5 minutes)?

openwisp_notifications/websockets/consumers.py
nemesisdesign: Have you tested the idea?

openwisp_notifications/websockets/handlers.py
```diff
@@ -1,8 +1,28 @@
...
+        in_notification_storm = True
+    if in_notification_storm:
+        cache.set(f'ow-noti-storm-{user.pk}', True, 60)
+    return in_notification_storm
```
nemesisdesign, line 24: Now that it's working better, I think we can centralize these numbers in a dictionary and allow this dict to be overridden via settings if needed, what do you think?

nemesisdesign (changes requested):
Great progress @pandafy!

After the sleep of 20 seconds in the test code example you shared I hear the sound again, altough only once, is that intentional or not? It's not clear how this works to me now so I am not sure whether this is a bug or a feature and whether we can keep it or we should change it, can you please clarify?

See also my comments below.

pandafy:
> After the sleep of 20 seconds in the test code example you shared I hear the sound again, altough only once, is that intentional or not? It's not clear how this works to me now so I am not sure whether this is a bug or a feature and whether we can keep it or we should change it, can you please clarify?

It was a bug. The time delay was of 20 seconds was perfect to trigger it since we reset the time. I have fixed it.

README.rst
```diff
@@ -776,6 +776,31 @@ model, then configure the setting as following:
...
 
+``OPENWISP_NOTIFICATIONS_NOTIFICATION_STORM_PREVENTION``
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+A ``dict`` containing configuration for notification storm prevention mechanism.
+The default configuration is as follows:
+
+.. code-block:: python
+
+    OPENWISP_NOTIFICATIONS_NOTIFICATION_STORM_PREVENTION = {
```
pandafy, line 787: We might need to re-write this documentation. I couldn't explain it really well 😓

openwisp_notifications/websockets/consumers.py
pandafy: https://github.com/openwisp/openwisp-notifications/issues/210#issue-1066359068

openwisp_notifications/settings.py
```diff
@@ -20,6 +20,19 @@
 POPULATE_PREFERENCES_ON_MIGRATE = getattr(
     settings, 'OPENWISP_NOTIFICATIONS_POPULATE_PREFERENCES_ON_MIGRATE', True
 )
+NOTIFICATION_STORM_PREVENTION = getattr(
+    settings,
+    'OPENWISP_NOTIFICATIONS_NOTIFICATION_STORM_PREVENTION',
+    {
+        'short_term_time_period': 10,
+        'short_term_notification_count': 6,
+        'long_term_time_period': 60,
```
nemesisdesign, line 29: so what about increasing this? 3 minutes to me sounds reasonable

nemesisdesign (changes requested):
There's a failing test, can you please check?

nemesisdesign (changes requested):
During testing I spotted this:

Screenshot from 2021-12-01 15-34-28

Can you please investigate?

README.rst
```diff
@@ -776,6 +776,31 @@ model, then configure the setting as following:
 This setting allows to disable creating `notification preferences <#notification-preferences>`_
 on running migrations.
 
+``OPENWISP_NOTIFICATIONS_NOTIFICATION_STORM_PREVENTION``
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+A ``dict`` containing configuration for notification storm prevention mechanism.
```
nemesisdesign, line 782: I think we have to explain what happens, eg:

When the system starts creating a lot of notifications because of a general network outage (eg: a black out), the notification storm prevention mechanism avoids to constantly display many notifications and ring the notification sound. 
The following configuration setting allows to tweak how this mechanism works, if needed.

Feel free to edit/improve.

openwisp_notifications/websockets/consumers.py
nemesisdesign: Ok, let's work on that soonish.

openwisp_notifications/websockets/consumers.py
```diff
@@ -37,20 +47,45 @@ def disconnect(self, close_code):
...
+                seconds=self._initial_backoff
+            ):
+                self.scope['last_update_datetime'] = datetime_now + timedelta(
+                    seconds=self.scope['backoff']
                 )
+                self.scope['backoff'] = self.scope['backoff'] + self._backoff_increment
+                event['reload_widget'] = False
+            else:
+                self.scope['last_update_datetime'] = datetime_now
+                self.scope['backoff'] = self._initial_backoff
```
nemesisdesign, line 78: is it possible to move this logic into a dedicated method?

openwisp_notifications/websockets/handlers.py
```diff
@@ -1,7 +1,55 @@
...
+from ..swapper import load_model
+
+Notification = load_model('Notification')
+
+
+def user_in_notification_storm(user):
+    """
+    A user is affected by notifications storm if it receives
+     - more than 6 notifications in 10 seconds
+     - more than 30 notifications in 60 seconds
```
nemesisdesign, line 19: this docstring may have to be updated, I think it may be good to explain it in a more general way, since the defaults can be changed in the future and this docstring will remain outdated.

nemesisdesign (changes requested):
See my comments below.

nemesisdesign (approved):
:+1:

PR #212, opened 2021-12-03
[fix] Fixed closing notification toast on slow connections #211

### LINKED ISSUE #211: [bug] Notification toast does not close on slow internet connections
The frontend performs two operation upon receiving a notification on websocket:
1. It displays the notification 
2. It reloads the notification widget

On a slow network, it can take time to reload the widget while the notification toast is already being displayed (notification content is sent via WS). 

Clicking on the close icon of notification toast not only closes the notification toast but also flags the notification as read(by removing the `unread` class of the notification element in notification widget and updating the backend via WS). Performing above observation in a state where notification widget is still reloading yields this on console:

Screenshot from 2021-12-03 22-20-20

As a result, the notification toast does not get closed. 

Following code is responsible for flagging notification as read. 

https://github.com/openwisp/openwisp-notifications/blob/67f344b04132676f16e4485388a68966ab8192c0/openwisp_notifications/static/openwisp-notifications/js/notifications.js#L324-L338

**Possible Fix**
Remove the `throw error` statement form the catch clause.

pandafy:
Notification toast will close and flag notification as read
even if notification widget is still reloading.

Closes #211

### DIFF
--- a/openwisp_notifications/static/openwisp-notifications/js/notifications.js
+++ b/openwisp_notifications/static/openwisp-notifications/js/notifications.js
@@ -323,18 +323,18 @@ function notificationWidget($) {
 
 function markNotificationRead(elem) {
     let elemId = elem.id.replace('ow-', '');
+    try {
+        document.querySelector(`#${elem.id}.ow-notification-elem`).classList.remove('unread');
+    } catch (error) {
+        // no op
+    }
+    notificationReadStatus.set(elemId, 'read');
     notificationSocket.send(
         JSON.stringify({
             type: 'notification',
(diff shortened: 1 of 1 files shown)


pandafy:
Before:

https://user-images.githubusercontent.com/32094433/144645726-8d5b5d33-6ac3-4ced-8358-b8d1709b8f8f.mp4



After:

https://user-images.githubusercontent.com/32094433/144645745-257b8c22-3e04-423b-b304-26df70eb1ee0.mp4

nemesisdesign (approved):
:+1:

PR #216, opened 2022-01-11
[fix] Fixed NotificationReadAllView 

pandafy:
- Fixed browsable API view for NotificationReadAllView
- Moved updating database query to celery worker

### DIFF
--- a/openwisp_notifications/api/views.py
+++ b/openwisp_notifications/api/views.py
@@ -99,8 +99,9 @@ class NotificationReadRedirect(BaseNotificationView):
 class NotificationReadAllView(BaseNotificationView):
     def post(self, request, *args, **kwargs):
         queryset = self.get_queryset()
-        queryset.update(unread=False)
-        # update() does not create post_save signal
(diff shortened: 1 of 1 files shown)


openwisp_notifications/tasks.py
```diff
@@ -201,3 +201,8 @@ def delete_ignore_object_notification(instance_id):
     Deletes IgnoreObjectNotification object post it's expiration.
     """
     IgnoreObjectNotification.objects.filter(id=instance_id).delete()
+
+
+@shared_task(soft_time_limit=600, time_limit=1800)
+def flag_user_notifications_as_read(user_id):
+    Notification.objects.filter(recipient_id=user_id).update(unread=False)
```
nemesisdesign, line 208: let's filter by unread here

nemesisdesign (changes requested):

nemesisdesign (changes requested):
Let's try bringing it back to the foreground but keep the improved query (that is, filter by unread).

nemesisdesign (approved):
Thanks.

PR #221, opened 2022-02-03
[fix] Notification when disabled during developement #159

### LINKED ISSUE #159: [fix] Receives notification even when disabled during development
**Current Behavior**
Disable notifications during development and then run `manage.py create_notification` command. You will get a notification even when you have disabled it.

**Expected Behavior**
You should not receive any notification. 

**Root Cause**
https://github.com/openwisp/openwisp-notifications/blob/19bf2ed59b11507d6b025e67fbd6973d4f83673f/openwisp_notifications/management/commands/__init__.py#L12-L15
`default_org` is passed as target here and it does not have `organization_id` in below code so this is causing the bug. 
https://github.com/openwisp/openwisp-notifications/blob/19bf2ed59b11507d6b025e67fbd6973d4f83673f/openwisp_notifications/handlers.py#L64

By changing `organization_id` to `id` it is solving the bug. But this is not a proper solution so we need to find a better solution for it.

Aryamanz29:
Fixed receiving Notification even when disabled during development

Fixes #159

### DIFF
--- a/openwisp_notifications/management/commands/__init__.py
+++ b/openwisp_notifications/management/commands/__init__.py
@@ -12,6 +12,7 @@ Organization = swapper_load_model('openwisp_users', 'Organization')
 class BaseCreateNotificationCommand(BaseCommand):
     def handle(self, *args, **kwargs):
         default_org = Organization.objects.first()
+        default_org.organization_id = default_org.id
         notify.send(sender=default_org, type='default', target=default_org)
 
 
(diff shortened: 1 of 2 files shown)


Aryamanz29:
> @Aryamanz29 you need to make changes in test too for this. A test case that fails before this fix should be added

Sure, I'll add test case

Aryamanz29:
> @Aryamanz29 you need to make changes in test too for this. A test case that fails before this fix should be added

PR now ready for review/testing :+1:  @devkapilbansal

nemesisdesign (approved):
Thanks :+1:

PR #222, opened 2022-02-03
[Fix] Error handling send email notification #179

### LINKED ISSUE #179: [bug] Add error handling in send_email_notification handler

aagman945:
Added exception handling for the issue #179 and test for the same . So ,If throws exception the function will be returned empty and mail will not be sent.

### DIFF
--- a/openwisp_notifications/tests/test_notifications.py
+++ b/openwisp_notifications/tests/test_notifications.py
@@ -818,6 +818,7 @@ class TestNotifications(TestOrganizationMixin, TransactionTestCase):
         data = dict(
             email_subject='Test Email subject', url='https://localhost:8000/admin'
         )
+        self.admin.notificationsetting_set.all().delete()
         Notification.objects.create(
             actor=self.admin,
             recipient=self.admin,
@@ -828,7 +829,7 @@ class TestNotifications(TestOrganizationMixin, TransactionTestCase):
             data=data,
             type="default",
         )
-        self.assertEqual(len(mail.outbox), 1)
+        self.assertEqual(len(mail.outbox), 0)
 
 
 class TestTransactionNotifications(TestOrganizationMixin, TransactionTestCase):
--- a/openwisp_notifications/tests/test_notifications.py
+++ b/openwisp_notifications/tests/test_notifications.py
@@ -813,7 +813,6 @@ class TestNotifications(TestOrganizationMixin, TransactionTestCase):
         user.delete()
         mocked_task.assert_not_called()
 
-
     def test_email_notif_without_notif_setting(self):
         target_obj = self._get_org_user()
         data = dict(
@@ -827,9 +826,10 @@ class TestNotifications(TestOrganizationMixin, TransactionTestCase):
             action_object=target_obj,
             target=target_obj,
             data=data,
-            type="default"
+            type="default",
         )
         self.assertEqual(len(mail.outbox), 1)
+
 
 class TestTransactionNotifications(TestOrganizationMixin, TransactionTestCase):
     def setUp(self):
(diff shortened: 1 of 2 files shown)


openwisp_notifications/tests/test_notifications.py
```diff
@@ -851,3 +852,28 @@ def test_notification_cache_update(self):
...
         self.assertEqual(operator_cache.username, 'new operator name')
+
+class test_email_notification(TestOrganizationMixin, TransactionTestCase):
+    def setUp(self):
+        self.user=self._create_user()
+        self.admin = self._create_admin()
+        self.target_obj = self._get_org_user()
+        self.data = dict(
+            email_subject='Test Email subject', url='https://localhost:8000/admin'
+        )
```
pandafy, line 863: We don't need to add a dedicated test class here. You can add this test to `TestTransactionNotifications` class. 

P.S.: The python convention is to use PascalCase for naming classes not snake_case. Refer https://namingconvention.org/python/

pandafy (changes requested):
@aagman945 kindly **read the contributing guidelines** and update your editor settings as mentioned here https://openwisp.io/docs/developer/contributing.html#avoiding-unnecessary-changes. 

Doing so would have reduced requested changes in this PR. The commit message needs to be updated as well. 

Also, always run `./run-qa-check` before pushing changes to GitHub.

aagman945:
@pandafy ./run-qa-checks throws error 
TypeError: [files].flat is not a function
at standalone (/usr/local/lib/node_modules/stylelint/lib/standalone.js:158:25)
at module.exports (/usr/local/lib/node_modules/stylelint/lib/cli.js:484:9)
at Object.<anonymous> (/usr/local/lib/node_modules/stylelint/bin/stylelint.js:8:22)
at Module._compile (internal/modules/cjs/loader.js:778:30)
at Object.Module._extensions..js (internal/modules/cjs/loader.js:789:10)
at Module.load (internal/modules/cjs/loader.js:653:32)
at tryModuleLoad (internal/modules/cjs/loader.js:593:12)
at Function.Module._load (internal/modules/cjs/loader.js:585:3)
at Function.Module.runMain (internal/modules/cjs/loader.js:831:12)
at startup (internal/bootstrap/node.js:283:19)

openwisp_notifications/tests/test_notifications.py
```diff
@@ -813,6 +813,23 @@ def test_delete_obsolete_tasks(self, mocked_task, *args):
...
+            recipient=self.admin,
+            description='Test Notification Description',
+            verb='Test Notification',
+            action_object=target_obj,
+            target=target_obj,
+            data=data,
+            type="default",
+        )
+        self.assertEqual(len(mail.outbox), 1)
+
```
pandafy, line 832: @aagman945 this test does not trigger the bug. The bug only appears if notification setting is not present for the  user. 

In the end of the test, you are assert for presence of one email. But we don't want an email to be sent in this scenario. This depicts that the test does not tests the exception handling that  you just added. 

I would change this test as following:

```suggestion
def test_email_notification_without_notification_setting(self):
        target_obj = self._get_org_user()
        data = dict(
            email_subject='Test Email subject', url='https://localhost:8000/admin'
        )
        # Delete notification settings of the recipient
        self.admin.notificationsetting_set.all().delete()
        Notification.objects.create(
            actor=self.admin,
            recipient=self.admin,
            description='Test Notification Description',
            verb='Test Notification',
            action_object=target_obj,
            target=target_obj,
            data=data,
            type="default",
        )
        self.assertEqual(len(mail.outbox), 0)
```

pandafy (changes requested):

pandafy (approved):

PR #223, opened 2022-02-04
[fix] Project dependent test cases #157

### LINKED ISSUE #157: [bug] Fix project dependent test cases
The following test checks whether a notification preference has been made for the default notification type. In other modules of openwisp(openwisp-controller) we unregister this notification type. As a consequence, this test fails when the test suite is run outside of the test project of openwisp-notifications. 

https://github.com/openwisp/openwisp-notifications/blob/19bf2ed59b11507d6b025e67fbd6973d4f83673f/openwisp_notifications/tests/test_admin.py#L200-L206

One solution to this could be to check for the default organization instead of checking for default notification type.

Aryamanz29:
Check for the default organization instead of default notification type,
Fixes #157

### DIFF
--- a/openwisp_notifications/tests/test_admin.py
+++ b/openwisp_notifications/tests/test_admin.py
@@ -23,6 +23,7 @@ Notification = load_model('Notification')
 NotificationSetting = load_model('NotificationSetting')
 notification_queryset = Notification.objects.order_by('-timestamp')
 Group = swapper_load_model('openwisp_users', 'Group')
+Organization = swapper_load_model('openwisp_users', 'Organization')
 
 
 class MockUser:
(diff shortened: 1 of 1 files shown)


openwisp_notifications/tests/test_admin.py
```diff
@@ -205,7 +205,7 @@ def test_notification_setting_inline_organization_formfield(self):
         response = self.client.get(
             reverse('admin:openwisp_users_user_change', args=(self.admin.pk,))
         )
-        self.assertContains(response, '<option value="default">Default Type</option>')
+        self.assertContains(response, 'default</option>')
```
Aryamanz29, line 208: (This  `'default</option>'` works as well and also saves one db query):

Alternate option :point_down: 
```python
Organization = swapper_load_model('openwisp_users', 'Organization')
....
....
....
self.assertContains(response, '<option value="{Organization.objects.first().id}">default</option>')
```

pandafy: I think this way we are making the test fragile. We should atleast check for organization id in the HTML response.

pandafy (changes requested):

openwisp_notifications/tests/test_admin.py
Aryamanz29: > I think this way we are making the test fragile. We should atleast check for organization id in the HTML response.

So should I go with this  method? @pandafy  👇
```py
Organization = swapper_load_model('openwisp_users', 'Organization')
....
....
....
self.assertContains(response, '<option value="{Organization.objects.first().id}">default</option>')
```

pandafy (approved):

PR #224, opened 2022-02-14
[change] Added celery time limits to tasks #217 

### LINKED ISSUE #217: [change] Add celery time limits to tasks
Depends on https://github.com/openwisp/openwisp-utils/issues/271.

Let's ensure all tasks have time limits defined. For similar tasks we may share the same settings.

pandafy:
Closes #217

### DIFF
--- a/openwisp_notifications/tasks.py
+++ b/openwisp_notifications/tasks.py
@@ -9,6 +9,7 @@ from django.utils import timezone
 
 from openwisp_notifications.swapper import load_model, swapper_load_model
 from openwisp_notifications.types import NOTIFICATION_TYPES
+from openwisp_utils.tasks import OpenwispCeleryTask
 
 User = get_user_model()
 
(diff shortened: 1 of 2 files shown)


openwisp_notifications/tasks.py
```diff
@@ -56,12 +57,12 @@ def delete_obsolete_objects(instance_app_label, instance_model, instance_id):
...
 
 
-@shared_task
+@shared_task(base=OpenwispCeleryTask)
 def delete_notification(notification_id):
     Notification.objects.filter(pk=notification_id).delete()
 
 
-@shared_task
+@shared_task(base=OpenwispCeleryTask)
```
nemesisdesign, line 65: this task could definitely take more than 10 seconds in some cases, I think the default timeouts are too strict, looking at this I would be even more conservative and increase the hard time limit to 2 minutes instead of 1 as I suggested in my other review.

nemesisdesign (changes requested):

requirements.txt
```diff
@@ -3,7 +3,6 @@ channels~=3.0.2
 # TODO set the next point release of openwisp-users when ready
 openwisp-users @ https://github.com/openwisp/openwisp-users/tarball/master
 # TODO set the next point release of openwisp-utils when ready
-openwisp-utils[rest] @ https://github.com/openwisp/openwisp-utils/tarball/master
+openwisp-utils[rest,celery] @ https://github.com/openwisp/openwisp-utils/tarball/issues/271-celery-task
```
nemesisdesign, line 6: requirements can be updated now

nemesisdesign (changes requested):

nemesisdesign (approved):

PR #231, opened 2022-04-11
[change!] Made API url structure consistent with other OpenWISP modules #230

### LINKED ISSUE #230: [change!] API URL structure is not consistent with other OpenWISP modules
As in OpenWISP Users, where we added the `/users/` prefix to all URLs, here we should add the `/notifications/` prefix (name of the module without `openwisp-`), so `/api/v1/notification/` would become `/api/v1/notifications/notification/` and so on.

This can have a negative impact on the UI and OpenWISP Controller, so please let's ensure those continue working as usual after this change.

codesankalp:
Closes #230

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -865,8 +865,8 @@ the results in conjunction with the ``page`` parameter.
 
 .. code-block:: text
 
-    GET /api/v1/notification/?page_size=10
-    GET /api/v1/notification/?page_size=10&page=2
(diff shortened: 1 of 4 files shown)


nemesisdesign (changes requested):
The README was not updated.

nemesisdesign (approved):

PR #242, opened 2022-06-10
[bug] Organization.DoesNotExist #238

### LINKED ISSUE #238: [bug] Organization.DoesNotExist   openwisp_notifications.tasks.ns_organization_created
When creating a new org in a prod system this exception pops up.

I think it's because we have to use `transaction.on_commit` on the handler as we use for the other receivers, otherwise the celery task is spawned before the transaction is saved to the DB and the organization does not exist yet.

yashikajotwani12:
Fixes #238

### DIFF
--- a/openwisp_notifications/handlers.py
+++ b/openwisp_notifications/handlers.py
@@ -274,11 +274,8 @@ def notification_type_registered_unregistered_handler(sender, **kwargs):
 def organization_user_post_save(instance, created, **kwargs):
     transaction.on_commit(
         lambda: tasks.update_org_user_notificationsetting.delay(
-            org_user_id=instance.pk,
-            user_id=instance.user_id,
-            org_id=instance.organization_id,
-            is_org_admin=instance.is_admin,
-        )
+            instance.pk, instance.user_id, instance.organization_id, instance.is_admin, created
+        )   
     )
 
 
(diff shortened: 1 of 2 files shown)


openwisp_notifications/handlers.py
```diff
@@ -305,9 +305,8 @@ def update_superuser_notification_settings(instance, created, **kwargs):
 @receiver(
     post_save, sender=Organization, dispatch_uid='org_created_notification_setting'
 )
-def notification_setting_org_created(created, instance, **kwargs):
-    if created:
-        tasks.ns_organization_created.delay(instance.pk)
+def notification_setting_org_created(instance, **kwargs):
+    transaction.on_commit(lambda: tasks.ns_organization_created.delay(instance.pk))
```
nemesisdesign, line 309: This logic must be executed only when an organization is created, otherwise it will be executed any time an organization is updated, which we do not want.

```suggestion
    if created:
        transaction.on_commit(lambda: tasks.ns_organization_created.delay(instance.pk))
```

nemesisdesign (changes requested):

atb00ker (approved):
LGTM! 😄

nemesisdesign (approved):

PR #259, opened 2023-04-09
[fix] Fixed `parse_header` error

keshav-04:
Recently, Django released its 4.2 version and remove the `django.http.multipartparser.parse_header()` function, So `djangorestframework` required version also get changed. So, added the new version of `djangorestframework` to use `django.utils.http.parse_header_parameters()` instead of `django.http.multipartparser.parse_header()`. As it is required by the notification widget. 

**Before**
image

**After**
image

### DIFF
--- a/requirements-test.txt
+++ b/requirements-test.txt
@@ -6,3 +6,4 @@ channels_redis~=3.4.0
 pytest-asyncio~=0.14.0
 pytest-django~=4.1.0
 freezegun~=1.1.0
+djangorestframework~=3.14.0


pandafy (commented):
@keshav-04 this appears as a Django and DRF dependency issue. We don't define version of those packages directly here. Those are defined in openwisp-utils and openwisp-users. I have created the [issue in openwisp-utils](https://github.com/openwisp/openwisp-utils/issues/337). We will track if from there. 

Please close this PR unless you think other changes are needed here.

PR #263, opened 2023-06-13
[deps] Added support for Django 4.1.0 and Django 4.2.0

pandafy:
- Dropped support for Python 3.7
- Bumped django-notifications-hq~=1.8.2
- Bumped markdown~=3.4.3


openwisp_notifications/migrations/0001_initial.py
pandafy, line 1: The latest version of django-notifications added verbose name to model fields. Since these does not affect the database schema, I updated the existing migration file.

nemesifier (approved):

Issue #249, opened 2022-07-11
[change] Allow relative paths


nemesifier:
It would be good to change the notification widget to allow the use of relative paths.
This helps the notification links to work also on systems which are using multiple domain names, some times a domain is reachable only from a private network but when users click on the notifications they are taken to the main domain name, which is not what they want.

BassCoder2808:
Hi @nemesisdesign, I wanted to contribute to this issue if it is still open, I have installed the openwisp-notifications and the setup part is done for the same, I wanted the steps like how can I reproduce the issue based on which I can change the code

arup1221:
@pandafy  I am Working on this issue.

arup1221:
> **Steps to replicate the issue**
> 
> 1. Run the test project in this repository
> 2. Edit the `/etc/hosts` file to point `internal.openwisp.test` and `external.openwisp.test` to `127.0.0.1` using the following command
> 
> ```
> sudo bash -c 'echo "127.0.0.1    internal.openwisp.test external.openwisp.test" >> /etc/hosts'
> ```
> 
> 3. Update the site objects at http://127.0.0.1:8000/admin/sites/site/ as following:
>    ![image](https://user-images.githubusercontent.com/32094433/231150656-1d0b32fe-99ad-4c3c-855e-7c59a01b347c.png)
>    **Note:** Instead of deleting the existing site object, edit it to use `external.openwisp.test`.
> 4. Generate a default notification by using the `create_notification` management command from the shell
> 
> ```
> cd tests/
> ./manage.py create_notification
> ```
> 
> 5. Open http://internal.openwisp.test:8000/admin/ and click on the notification in the notification widget
> 
> **Actual result** After clicking on the notification, the user to sent to the target object with `external.openwisp.test` domain. E.g.: `http://external.openwisp.test:8000/admin/openwisp_users/organization/b150de21-65b2-4499-b1a8-3af78e8442b5/change/`
> 
> **Expected result** After clicking on the notification, the URL should take the user to the target object with `internal.openwisp.test` domain. E.g.: `http://internal.openwisp.test:8000/admin/openwisp_users/organization/b150de21-65b2-4499-b1a8-3af78e8442b5/change/`

is this thing still working or any updation is made?

pandafy:
It is important to update the `ALLOWED_HOSTS` settings in tests/settings.py like below to replicate this issue

```python
[...]

PR #266, opened 2024-03-03
chore: relative url for notification links

Dhanus3133:
Fixes: #249 

image

### DIFF
--- a/openwisp_notifications/tests/test_notifications.py
+++ b/openwisp_notifications/tests/test_notifications.py
@@ -5,7 +5,6 @@ from celery.exceptions import OperationalError
 from django.apps.registry import apps
 from django.contrib.auth import get_user_model
 from django.contrib.contenttypes.models import ContentType
-from django.contrib.staticfiles.testing import StaticLiveServerTestCase
 from django.core import mail
 from django.core.cache import cache
 from django.core.exceptions import ImproperlyConfigured
@@ -15,9 +14,6 @@ from django.test import TransactionTestCase
 from django.urls import reverse
 from django.utils import timezone
 from django.utils.timesince import timesince
-from selenium.webdriver.common.by import By
-from selenium.webdriver.support import expected_conditions as EC
-from selenium.webdriver.support.ui import WebDriverWait
 
 from openwisp_notifications import settings as app_settings
 from openwisp_notifications import tasks
@@ -37,9 +33,8 @@ from openwisp_notifications.types import (
     _unregister_notification_choice,
     get_notification_configuration,
 )
-from openwisp_notifications.utils import _get_absolute_url, _get_object_link
+from openwisp_notifications.utils import _get_absolute_url
 from openwisp_users.tests.utils import TestOrganizationMixin
-from openwisp_utils.test_selenium_mixins import SeleniumTestMixin
 from openwisp_utils.tests import capture_any_output
 
 User = get_user_model()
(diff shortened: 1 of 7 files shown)


openwisp_notifications/base/models.py
```diff
@@ -84,22 +84,19 @@ def _get_related_object_url(self, field):
...
-    @property
-    def action_url(self):
-        return self._get_related_object_url(field='action_object')
+    def action_url(self, absolute_url=False):
+        return self._get_related_object_url(field='action_object', absolute_url=absolute_url)
 
-    @property
-    def target_url(self):
-        return self._get_related_object_url(field='target')
+    def target_url(self, absolute_url=False):
```
nemesifier, line 98: If we turn these properties into methods which support passing arguments we have to change their name to `get_{property_name}`, for backward compatibility we can keep the properties and internally call the new method which passes `absolute_url=True`.

nemesifier (changes requested):
Thank you @Dhanus3133, it looks good.

Here's a few things to improve, see below.

The REST API is returning relative URLs as shown in the following screenshot:

Screenshot from 2024-03-07 16-29-40

That's unintended, we only want this new behavior in the admin widget, the API may be hosted on a different domain and hence must return the full URL.

Please also add a new unit test which tests the new methods that allow returning the relative URLs.

Watch out also from build failures because of failed QA checks, please read the build output to find out how to run qa checks locally and how to fix common formatting and commit message format issues.

Dhanus3133:
Updated with just the required notification widget changes, although the QA checks for the CSS files are failing for some reason.

nemesifier (changes requested):
Works as expected now, thanks!
Take a look at the comments below for some minor code improvements.

If you can add a selenium browser test for this it would be great, let me know!

Dhanus3133:
@nemesifier, I tried adding Selenium tests, but when the notification tab is clicked, it just loads. When I look into the console, it says it cannot find WS and returns 404 in the network.
image
Could you help me here? Thanks.

nemesifier:
> it just loads

Did you mean "it doesn't load"?

Dhanus3133:
> Did you mean "it doesn't load"?

I meant the WebSocket connection is not found, so in the front end, it just waits, causing the spinning UI.

https://github.com/openwisp/openwisp-notifications/assets/43700516/1c21bb68-f2ec-4c4c-9fb5-da85fe4b112e

nemesifier (changes requested):
The base test class `StaticLiveServerTestCase` does not support websockets.

Here's a rough outline of how we might structure these tests:

```python
from django.contrib.staticfiles.testing import StaticLiveServerTestCase
from channels.testing import ChannelsLiveServerTestCase

class SeleniumTestNotifications(StaticLiveServerTestCase, ChannelsLiveServerTestCase):
    # ... etc ...
```

Please consults these links for more information:

- [ChannelsLiveServerTestCase docs](https://channels.readthedocs.io/en/latest/topics/testing.html) 
- [StaticLiveServerTestCase docs](https://docs.djangoproject.com/en/5.0/ref/contrib/staticfiles/#specialized-test-case-to-support-live-testing)

nemesifier (commented):
@Dhanus3133 I missed the latest updates on this, will review asap.

openwisp_notifications/tests/test_notifications.py
```diff
@@ -927,3 +932,48 @@ def test_notification_cache_update(self):
         self.assertEqual(notification.target.username, 'new operator name')
         # Done for populating cache
         self.assertEqual(operator_cache.username, 'new operator name')
+
+
+class SeleniumTestNotifications(
```
nemesifier, line 937: I would move this to a new dedicated test file, it's possible that over time we'll be adding more tests for the widget.

requirements-test.txt
```diff
@@ -5,3 +5,4 @@ channels_redis~=4.1.0
 pytest-asyncio~=0.21.0
 pytest-django~=4.5.0
 freezegun~=1.2.2
+selenium~=4.18.1
```
nemesifier, line 8: We already depend on selenium in openwisp-utils[selenium], we may want to look if we can avoid redefining this here: https://github.com/openwisp/openwisp-utils/blob/master/setup.py#L67

tests/openwisp2/settings.py
```diff
@@ -14,6 +14,9 @@
     'default': {
         'ENGINE': 'django.db.backends.sqlite3',
         'NAME': os.path.join(BASE_DIR, 'openwisp_notifications.db'),
+        "TEST": {
+            "NAME": os.path.join(BASE_DIR, "openwisp_notifications_test.db"),
+        },
```
nemesifier, line 19: Are you sure this is really needed?

nemesifier (commented):
Great work @Dhanus3133, I left some comments below, I am doing some quick tests locally  to verify some of my comments.

nemesifier (approved):
@Dhanus3133 I pushed some improvements, feel free to review the changes. Once the build passes I will merge. :+1:

PR #267, opened 2024-03-03
[change] Allow relative paths #249

### LINKED ISSUE #249: [change] Allow relative paths

arup1221:
closes #249


arup1221:
Screenshot 2024-03-03 at 5 15 27 PM
Screenshot 2024-03-03 at 5 16 32 PM

nemesifier:
Thanks for trying!

Issue #254, opened 2022-08-17
[feature] Add dedicated notification type for internal errors


nemesifier:
Some times errors may happen in the background, at the moment we do not have a way to show these errors to users.

We could define a notification type for these situations and implement something slightly different in the UI: when clicking on such a notification, instead of being redirected to the target_url, an overlay can be opened with the long description of the error.

The overlay then can show a button which will direct the user to the target object.

BassCoder2808:
Hi @nemesisdesign, I wanted to contribute to this issue, I have installed the openwisp-notifications and the setup part is done for the same, I wanted to get a more brief understanding of this, like what exactly is the indented behavior and the steps to reproduce the issue in the background based on which we can notify the user

sawaisinghh:
Hey @nemesisdesign, I might be able to fix this issue. can you please assign it to me?

pandafy:
Email notifications should be disabled for this notification type.

PR #272, opened 2024-05-14
[feature] Description of notification shown in a dialog box

Dhanus3133:
Fixes: #254 

(Test fails for the new type)

### DIFF
--- a/openwisp_notifications/templates/admin/base_site.html
+++ b/openwisp_notifications/templates/admin/base_site.html
@@ -34,6 +34,17 @@
     <div class="ow-no-notifications ow-round-bottom-border ow-hide">
         <p>{% trans 'No new notification.' %}</p>
     </div>
+    <div class="ow-dialog-overlay ow-hide">
+        <div class="ow-dialog-notifications">
+            <span class="ow-dialog-close ow-dialog-close-x">&times;</span>
+            <h2 class="ow-message-title"></h2>
+            <div class="ow-message-description"></div>
+            <div class="ow-dialog-buttons">
+              <button class="ow-message-target-redirect ow-hide">Click to Redirect</button>
+              <button class="ow-dialog-close">Close</button>
+            </div>
+        </div>
+    </div>
   </div>
 {% endblock  %}
 
--- a/openwisp_notifications/api/serializers.py
+++ b/openwisp_notifications/api/serializers.py
@@ -59,6 +59,7 @@ class NotificationListSerializer(NotificationSerializer):
         fields = [
             'id',
             'message',
+            'description',
             'unread',
             'target_url',
             'email_subject',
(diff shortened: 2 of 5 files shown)


openwisp_notifications/templates/admin/base_site.html
```diff
@@ -34,6 +34,17 @@
...
         <p>{% trans 'No new notification.' %}</p>
     </div>
+    <div class="ow-dialog-overlay ow-hide">
+        <div class="ow-dialog-notifications">
+            <span class="ow-dialog-close ow-dialog-close-x">&times;</span>
+            <h2 class="ow-message-title"></h2>
+            <div class="ow-message-description"></div>
+            <div class="ow-dialog-buttons">
+              <button class="ow-message-target-redirect ow-hide">Click to Redirect</button>
+              <button class="ow-dialog-close">Close</button>
```
nemesifier, line 44: Always flag strings as translatable:
```suggestion
              <button class="ow-dialog-close">{% trans 'Close' %}</button>
```

openwisp_notifications/templates/admin/base_site.html
```diff
@@ -34,6 +34,17 @@
     <div class="ow-no-notifications ow-round-bottom-border ow-hide">
         <p>{% trans 'No new notification.' %}</p>
     </div>
+    <div class="ow-dialog-overlay ow-hide">
+        <div class="ow-dialog-notifications">
+            <span class="ow-dialog-close ow-dialog-close-x">&times;</span>
+            <h2 class="ow-message-title"></h2>
+            <div class="ow-message-description"></div>
+            <div class="ow-dialog-buttons">
+              <button class="ow-message-target-redirect ow-hide">Click to Redirect</button>
```
nemesifier, line 43: Same here (flag strings as translatable), however, I am not sure about "Click to Redirect", maybe we can just call it "Open"?

nemesifier (changes requested):
Good progress @Dhanus3133, see my comments below.

The tests which are failing are for the sample app, an example extension of this module, see: https://github.com/openwisp/openwisp-notifications/blob/master/tests/openwisp2/settings.py#L199-L219.

openwisp_notifications/tests/test_api.py
```diff
@@ -533,7 +533,7 @@ def test_notification_setting_list_api(self):
             self.assertEqual(len(response.data['results']), number_of_settings)
 
         with self.subTest('Test "page_size" query'):
-            page_size = 1
+            page_size = 2
```
pandafy, line 536: This test assumes that only one NotificationSetting to be present in the database which was correct when we only had the "default" notification type. We introduced another notification type, so there will be two NotificationSettings in the DB. 

When we set the page_size to 1 in the request, the response would contain only one object per page. Since, there are two objects now in the DB, there will be two pages available for this page size. Thus, the next URL is not empty. 

Read the[ DRF docs on pagination for more details](https://www.django-rest-framework.org/api-guide/pagination/).

Issue #270, opened 2024-04-03
[change] Do not send email notifications if the email of the user is not flagged as verified


nemesifier:
Do not send email notifications if the email of the user is not flagged as verified.

PR #274, opened 2024-05-15
[change] Do not send email notifications to unverified email addresses

Dhanus3133:
Closes: #270

### DIFF
--- a/openwisp_notifications/handlers.py
+++ b/openwisp_notifications/handlers.py
@@ -1,5 +1,6 @@
 import logging
 
+from allauth.account.models import EmailAddress
 from celery.exceptions import OperationalError
 from django.contrib.auth import get_user_model
 from django.contrib.contenttypes.models import ContentType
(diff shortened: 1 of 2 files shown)


openwisp_notifications/handlers.py
```diff
@@ -182,7 +183,12 @@ def send_email_notification(sender, instance, created, **kwargs):
         # therefore send email anyway.
         email_preference = True
 
-    if not (email_preference and instance.recipient.email):
+    email_address = EmailAddress.objects.filter(
+        user=instance.recipient, email=instance.recipient.email
+    ).first()
+    is_email_verified = email_address.verified if email_address else False
```
nemesifier, line 189: I think this can be shortened to just:

```
email_verified = recipient.emailaddress_set.filter(verified=True, email=recipient.email).exists()
```

nemesifier (changes requested):
Thanks @Dhanus3133, see my comments below.

A test is missing from this patch.

Please add a test that sends a notification to an user having an unverified email address and expects zero email sent.
The test should fail on the current master but pass on this branch.

openwisp_notifications/handlers.py
Dhanus3133: Updated with the changes.

nemesifier (approved):

PR #275, opened 2024-05-18
[feature] Added generic message notification type (shown in dialg box)

### LINKED ISSUE #254: [feature] Add dedicated notification type for internal errors

Dhanus3133:
Closes #254.

### DIFF
--- a/openwisp_notifications/api/serializers.py
+++ b/openwisp_notifications/api/serializers.py
@@ -59,7 +59,7 @@ class NotificationListSerializer(NotificationSerializer):
         fields = [
             'id',
             'message',
-            'get_description',
+            'rendered_description',
             'unread',
             'target_url',
             'email_subject',
--- a/openwisp_notifications/api/serializers.py
+++ b/openwisp_notifications/api/serializers.py
@@ -59,6 +59,7 @@ class NotificationListSerializer(NotificationSerializer):
         fields = [
             'id',
             'message',
+            'description',
             'unread',
             'target_url',
             'email_subject',
--- a/openwisp_notifications/api/serializers.py
+++ b/openwisp_notifications/api/serializers.py
@@ -59,7 +59,7 @@ class NotificationListSerializer(NotificationSerializer):
         fields = [
             'id',
             'message',
-            'description',
+            'get_description',
             'unread',
             'target_url',
             'email_subject',
--- a/openwisp_notifications/api/serializers.py
+++ b/openwisp_notifications/api/serializers.py
@@ -55,11 +55,13 @@ class NotificationSerializer(serializers.ModelSerializer):
 
 
 class NotificationListSerializer(NotificationSerializer):
+    description = serializers.CharField(source='rendered_description')
+
     class Meta(NotificationSerializer.Meta):
         fields = [
             'id',
             'message',
-            'rendered_description',
+            'description',
             'unread',
             'target_url',
             'email_subject',
--- a/openwisp_notifications/base/models.py
+++ b/openwisp_notifications/base/models.py
@@ -120,7 +120,9 @@ class AbstractNotification(UUIDModel, BaseNotification):
             try:
                 config = get_notification_configuration(self.type)
                 data = self.data or {}
-                if 'message' in config:
+                if 'message' in data:
+                    md_text = data['message']
+                elif 'message' in config:
                     md_text = config['message'].format(notification=self, **data)
                 else:
                     md_text = render_to_string(
--- a/openwisp_notifications/base/models.py
+++ b/openwisp_notifications/base/models.py
@@ -107,7 +107,7 @@ class AbstractNotification(UUIDModel, BaseNotification):
 
     @cached_property
     def rendered_description(self):
-        return mark_safe(markdown(self.description))
+        return mark_safe(markdown(self.description)) if self.description else None
 
     @property
     def email_message(self):
--- a/openwisp_notifications/base/models.py
+++ b/openwisp_notifications/base/models.py
@@ -106,7 +106,7 @@ class AbstractNotification(UUIDModel, BaseNotification):
         return self.get_message()
 
     @cached_property
-    def get_description(self):
+    def rendered_description(self):
         return mark_safe(markdown(self.description))
 
     @property
--- a/openwisp_notifications/base/models.py
+++ b/openwisp_notifications/base/models.py
@@ -107,40 +107,72 @@ class AbstractNotification(UUIDModel, BaseNotification):
 
     @cached_property
     def rendered_description(self):
-        return mark_safe(markdown(self.description)) if self.description else None
+        if not self.description:
+            return
+        md_description = self._get_formatted_string(self.description)
+        return mark_safe(markdown(md_description))
 
     @property
     def email_message(self):
         return self.get_message(email_message=True)
 
+    def _set_extra_attributes(self, email_message=False):
+        """
+        This method sets extra attributes for the notification
+        which are required to render the notification correctly.
+        """
+        self.actor_link = self.actor_url
+        self.action_link = self.action_url
+        self.target_link = (
+            self.target_url if not email_message else self.redirect_view_url
+        )
+
+    def _delete_extra_attributes(self):
+        """
+        This method removes the extra_attributes set by
+        _set_extra_attributes
+        """
+        for attr in ('actor_link', 'action_link', 'target_link'):
+            delattr(self, attr)
+
+    def _get_formatted_string(self, string, email_message=False):
+        self._set_extra_attributes(email_message=email_message)
+        data = self.data or {}
+        try:
+            formatted_output = string.format(notification=self, **data)
+        except AttributeError as exception:
+            self._invalid_notification(
+                self.pk,
+                exception,
+                'Error encountered in rendering notification message',
+            )
+        self._delete_extra_attributes()
+        return formatted_output
+
     def get_message(self, email_message=False):
         if self.type:
-            # setting links in notification object for message rendering
-            self.actor_link = self.actor_url
-            self.action_link = self.action_url
-            self.target_link = (
-                self.target_url if not email_message else self.redirect_view_url
-            )
             try:
                 config = get_notification_configuration(self.type)
                 data = self.data or {}
                 if 'message' in data:
                     md_text = data['message']
                 elif 'message' in config:
-                    md_text = config['message'].format(notification=self, **data)
+                    md_text = config['message']
                 else:
+                    self._set_extra_attributes(email_message=email_message)
                     md_text = render_to_string(
                         config['message_template'], context=dict(notification=self)
                     ).strip()
-            except (AttributeError, KeyError, NotificationRenderException) as exception:
+                    self._delete_extra_attributes()
+            except (KeyError, NotificationRenderException) as exception:
                 self._invalid_notification(
                     self.pk,
                     exception,
                     'Error encountered in rendering notification message',
                 )
-            # clean up
-            self.actor_link = self.action_link = self.target_link = None
-            return mark_safe(markdown(md_text))
+            return mark_safe(
+                markdown(self._get_formatted_string(md_text, email_message))
+            )
         else:
             return self.description
 
--- a/openwisp_notifications/base/models.py
+++ b/openwisp_notifications/base/models.py
@@ -105,6 +105,10 @@ class AbstractNotification(UUIDModel, BaseNotification):
     def message(self):
         return self.get_message()
 
+    @cached_property
+    def get_description(self):
+        return mark_safe(markdown(self.description))
+
     @property
     def email_message(self):
         return self.get_message(email_message=True)
--- a/openwisp_notifications/base/models.py
+++ b/openwisp_notifications/base/models.py
@@ -1,4 +1,5 @@
 import logging
+from contextlib import contextmanager
 
 from django.conf import settings
 from django.contrib.contenttypes.fields import GenericForeignKey
@@ -27,6 +28,35 @@ from openwisp_notifications.utils import _get_absolute_url, _get_object_link
 from openwisp_utils.base import UUIDModel
 
 logger = logging.getLogger(__name__)
+
+
+@contextmanager
+def notification_render_attributes(obj, **attrs):
+    """
+    This context manager sets temporary attributes on
+    the notification object to allowing rendering of
+    notification.
+
+    It can only be used to set aliases of the existing attributes.
+    By default, it will set the following aliases:
+        - actor_link -> actor_url
+        - action_link -> action_url
+        - target_link -> target_url
+    """
+    defaults = {
+        'actor_link': 'actor_url',
+        'action_link': 'action_url',
+        'target_link': 'target_url',
+    }
+    defaults.update(attrs)
+
+    for target_attr, source_attr in defaults.items():
+        setattr(obj, target_attr, getattr(obj, source_attr))
+
+    yield obj
+
+    for attr in defaults.keys():
+        delattr(obj, attr)
 
 
 class AbstractNotification(UUIDModel, BaseNotification):
@@ -103,78 +133,44 @@ class AbstractNotification(UUIDModel, BaseNotification):
 
     @cached_property
     def message(self):
-        return self.get_message()
+        with notification_render_attributes(self):
+            return self.get_message()
 
     @cached_property
     def rendered_description(self):
         if not self.description:
             return
-        md_description = self._get_formatted_string(self.description)
-        return mark_safe(markdown(md_description))
+        with notification_render_attributes(self):
+            data = self.data or {}
+            desc = self.description.format(notification=self, **data)
+        return mark_safe(markdown(desc))
 
     @property
     def email_message(self):
-        return self.get_message(email_message=True)
+        with notification_render_attributes(self, target_link='redirect_view_url'):
+            return self.get_message()
 
-    def _set_extra_attributes(self, email_message=False):
-        """
-        This method sets extra attributes for the notification
-        which are required to render the notification correctly.
-        """
-        self.actor_link = self.actor_url
-        self.action_link = self.action_url
-        self.target_link = (
-            self.target_url if not email_message else self.redirect_view_url
-        )
-
-    def _delete_extra_attributes(self):
-        """
-        This method removes the extra_attributes set by
-        _set_extra_attributes
-        """
-        for attr in ('actor_link', 'action_link', 'target_link'):
-            delattr(self, attr)
-
-    def _get_formatted_string(self, string, email_message=False):
-        self._set_extra_attributes(email_message=email_message)
-        data = self.data or {}
+    def get_message(self):
+        if not self.type:
+            return self.description
         try:
-            formatted_output = string.format(notification=self, **data)
-        except AttributeError as exception:
+            config = get_notification_configuration(self.type)
+            data = self.data or {}
+            if 'message' in data:
+                md_text = data['message'].format(notification=self, **data)
+            elif 'message' in config:
+                md_text = config['message'].format(notification=self, **data)
+            else:
+                md_text = render_to_string(
+                    config['message_template'], context=dict(notification=self, **data)
+                ).strip()
+        except (AttributeError, KeyError, NotificationRenderException) as exception:
             self._invalid_notification(
                 self.pk,
                 exception,
                 'Error encountered in rendering notification message',
             )
-        self._delete_extra_attributes()
-        return formatted_output
-
-    def get_message(self, email_message=False):
-        if self.type:
-            try:
-                config = get_notification_configuration(self.type)
-                data = self.data or {}
-                if 'message' in data:
-                    md_text = data['message']
-                elif 'message' in config:
-                    md_text = config['message']
-                else:
-                    self._set_extra_attributes(email_message=email_message)
-                    md_text = render_to_string(
-                        config['message_template'], context=dict(notification=self)
-                    ).strip()
-                    self._delete_extra_attributes()
-            except (KeyError, NotificationRenderException) as exception:
-                self._invalid_notification(
-                    self.pk,
-                    exception,
-                    'Error encountered in rendering notification message',
-                )
-            return mark_safe(
-                markdown(self._get_formatted_string(md_text, email_message))
-            )
-        else:
-            return self.description
+        return mark_safe(markdown(md_text))
 
     @cached_property
     def email_subject(self):
--- a/openwisp_notifications/static/openwisp-notifications/css/notifications.css
+++ b/openwisp_notifications/static/openwisp-notifications/css/notifications.css
@@ -16,6 +16,7 @@
   justify-content: center;
   align-items: center;
   transition: opacity 0.3s;
+  z-index: 9999;
 }
 .ow-dialog-notification {
   position: relative;
--- a/openwisp_notifications/static/openwisp-notifications/css/notifications.css
+++ b/openwisp_notifications/static/openwisp-notifications/css/notifications.css
@@ -5,6 +5,68 @@
   border-bottom-left-radius: 3px;
   border-bottom-right-radius: 3px;
 }
+.ow-dialog-overlay {
+  position: fixed;
+  left: 0;
+  top: 0;
+  width: 100%;
+  height: 100%;
+  background-color: rgba(0, 0, 0, 0.6);
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  transition: opacity 0.3s;
+}
+.ow-dialog-notifications {
+  position: relative;
+  background-color: white;
+  padding: 20px;
+  padding-top: 40px;
+  border-radius: 10px;
+  box-shadow: 0 5px 15px rgba(0, 0, 0, 0.3);
+  width: 90%;
+  max-width: 500px;
+  text-align: left;
+}
+.ow-dialog-close-x {
+  color: #333;
+  cursor: pointer;
+  font-size: 1.25em;
+  position: absolute;
+  display: block;
+  font-weight: bold;
+  top: 10px;
+  right: 10px;
+}
+.ow-dialog-close-x:hover {
+  color: #e04343;
+}
+.ow-message-title {
+  color: #333;
+  margin-bottom: 10px;
+}
+.ow-message-title a {
+  color: #df5d43 !important;
+  font-weight: bold;
+}
+.ow-message-description {
+  margin-bottom: 20px;
+  line-height: 1.6;
+  color: #666;
+}
+.ow-dialog-buttons button {
+  margin-right: 10px;
+  background-color: #007bff;
+  color: white;
+  border: none;
+  padding: 10px 20px;
+  border-radius: 5px;
+  cursor: pointer;
+  transition: background-color 0.3s;
+}
+.ow-dialog-buttons button:hover {
+  background-color: #0056b3;
+}
 
 /* Notification bell */
 .ow-notifications {
--- a/openwisp_notifications/static/openwisp-notifications/css/notifications.css
+++ b/openwisp_notifications/static/openwisp-notifications/css/notifications.css
@@ -17,7 +17,7 @@
   align-items: center;
   transition: opacity 0.3s;
 }
-.ow-dialog-notifications {
+.ow-dialog-notification {
   position: relative;
   background-color: white;
   padding: 20px;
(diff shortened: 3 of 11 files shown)


nemesifier (changes requested):
@Dhanus3133 as part of UI changes, it's nice to post screenshots during reviews. Can you please add a few here?

I think we need to mention this new feature in the README, at the end of the `Notification Types` section.

We also need to add a basic test, we could reuse the test file we added recently, `test_widget`.

openwisp_notifications/tests/test_api.py
```diff
@@ -533,7 +533,7 @@ def test_notification_setting_list_api(self):
             self.assertEqual(len(response.data['results']), number_of_settings)
 
         with self.subTest('Test "page_size" query'):
-            page_size = 1
+            page_size = 2
```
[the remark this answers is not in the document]
Dhanus3133: > This test assumes that only one NotificationSetting to be present in the database which was correct when we only had the "default" notification type. We introduced another notification type, so there will be two NotificationSettings in the DB.
> 
> When we set the page_size to 1 in the request, the response would contain only one object per page. Since, there are two objects now in the DB, there will be two pages available for this page size. Thus, the next URL is not empty.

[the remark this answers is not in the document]
nemesifier: Got it, thanks.

openwisp_notifications/types.py
```diff
@@ -17,9 +17,26 @@
         'email_notification': True,
         'web_notification': True,
     },
+    'general_message': {
```
nemesifier, line 20: @pandafy @Dhanus3133 what do you think about changing this to just `generic`?

Rather than general, it's generic. This is a generic notification type that we are adding to allow any module to send information to users for any need that may arise.

The fact that it's a message it's also kind of implied, so for these reasons I would just flag it `generic` for the sake of brevity.

openwisp_notifications/types.py
```diff
@@ -17,9 +17,26 @@
...
+        'level': 'info',
+        'verb': 'message verb',
+        'verbose_name': 'Message Type',
+        'email_subject': '[{site.name}] Message Notification Subject',
+        'message': (
+            'Message notification with {notification.verb} and level {notification.level}'
+            ' by [{notification.actor}]({notification.actor_link})'
+        ),
+        'description': '{notification.description}',
+        'message_template': 'openwisp_notifications/default_message.md',
```
nemesifier, line 30: @pandafy: if we have `message_template` here, do we need to also define the `message` key above?
Isn't it redundant? If so, I would remove `message` as moreover it's not flagged as translatable.

nemesifier (changes requested):
One more note about the README, let's also add an example of usage of this notification type, eg: a few lines of code which send a generic notification.

nemesifier (changes requested):
A few more quick notes. Let's discuss these in our weekly meeting.

## Short summary

I try to create a test notification with:

```python
default_org = Organization.objects.first()
notify.send(
    sender=default_org,
    description='Test.',
    message='Something alarming happened!',
    type='general_message',
    target=default_org
)
```

I see a notification created with the following text:

``` 
Message notification with message verb and level info by default.
```

How can modules change this text to convey a short summary to the user?

## Links

Clicking on any link in the notification causes the dialog to appear but the page reloads and the dialog disappears, which is not good for UX.

Let's discuss possible solutions

## Not specifying a target causes JS to fail

Try the following.

```python
default_org = Organization.objects.first()
notify.send(
    sender=default_org,
    description='Test.',
    message='Something alarming happened!',
    type='general_message'
)```


The JS fails for me with the following error:

```
notifications.js:215 Uncaught TypeError: Failed to construct 'URL': Invalid URL
    at notificationListItem (notifications.js:215:28)
    at notifications.js:117:26
    at Array.forEach (<anonymous>)
    at pageContainer (notifications.js:116:14)
    at appendPage (notifications.js:124:46)
    at Object.success (notifications.js:162:21)
    at fire (jquery.js:3536:31)
    at Object.fireWith [as resolveWith] (jquery.js:3666:7)
    at done (jquery.js:9878:14)
    at XMLHttpRequest.<anonymous> (jquery.js:10139:9)
﻿```

If I am not mistaken, the goal here is to allow creating generic messages that may not point to any object if needed.

## Overlay issue

Screenshot from 2024-05-20 17-04-03

## Markdown not support

Screenshot from 2024-05-20 17-09-06
ed

I probably forgot to specify this in the issue desc, but our notification messages support markdown to allow us to format the text if needed.

Can you support markdown in the rendering of this overlay?

## Make X close button bigger

The X close button seems too small to me right now, can you please increase it slightly?

openwisp_notifications/types.py
pandafy: Yes, only one of them is required.

openwisp_notifications/static/openwisp-notifications/css/notifications.css
```diff
@@ -5,6 +5,68 @@
...
+  color: white;
+  border: none;
+  padding: 10px 20px;
+  border-radius: 5px;
+  cursor: pointer;
+  transition: background-color 0.3s;
+}
+.ow-dialog-buttons button:hover {
+  background-color: #0056b3;
+}
```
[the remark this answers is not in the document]
Dhanus3133: I have reduced most of the CSS like on the buttons, but the existing CSS was not very helpful on the overlays.

openwisp_notifications/api/serializers.py
```diff
@@ -59,6 +59,7 @@ class Meta(NotificationSerializer.Meta):
         fields = [
             'id',
             'message',
+            'get_description',
```
Dhanus3133, line 62: Is the naming like this fine here? Would like to get some suggestions on how it can be.

openwisp_notifications/tests/test_widget.py
```diff
@@ -54,3 +53,27 @@ def test_notification_relative_link(self):
...
+        self.notification_options.update(
+            {'message': 'Test Message', 'description': 'Test Description'}
+        )
+        notification = self._create_notification().pop()[1][0]
+        self.web_driver.find_element(By.ID, 'openwisp_notifications').click()
+        time.sleep(4)
+        WebDriverWait(self.web_driver, 10).until(
+            EC.visibility_of_element_located((By.ID, f'ow-{notification.id}'))
+        )
+        self.web_driver.find_element(By.ID, f'ow-{notification.id}').click()
```
Dhanus3133, line 68: The click on the notification element doesn't seem to be clicking properly. I'm not sure why this occurs, but when manually clicked, it has no issues.

Dhanus3133 (commented):
Need some suggestions here.

nemesifier (changes requested):
Good progress @Dhanus3133!

This review is not exhaustive, I just want to point out the most obvious issues I see after your changes.

I will do more testing in the coming days but in the meanwhile you can fix the issues below.

### Menu still not covered by overlay

This problem still stands:

Screenshot from 2024-05-23 20-01-37

I recommend adding the overlay programmatically in a different HTML root element to fix this.

### Notification shown on screen does not trigger overlay

Execute the following from the django shell while you have the application running in your browser:

```python
from openwisp_notifications.signals import notify
notify.send(
    sender=default_org,
    description='Test.',
    message='Something alarming happened!',
    type='generic_message',
    target=default_org
)
```

The following appears:

Screenshot from 2024-05-23 20-01-37

Try to click on it.
I would expect the generic message to appear, but that's not the case.

openwisp_notifications/static/openwisp-notifications/js/notifications.js
```diff
@@ -69,24 +72,50 @@ function initNotificationDropDown($) {
...
+        ) {
             // Don't hide if focus changes to notification bell icon
             if (e.target != $('#openwisp_notifications').get(0)) {
                 $('.ow-notification-dropdown').addClass('ow-hide');
             }
         }
     });
 
-    $('.ow-notification-dropdown').on('keyup', '*', function(e){
+    $(document).on('keyup', '*', function(e){
```
nemesifier, line 86: This will fire on every keyup event that will happen in OpenWISP. Are you sure this change is intended?
Can't we narrow down the firing a bit to make sure it's not fired unless needed?

openwisp_notifications/static/openwisp-notifications/js/notifications.js
```diff
@@ -324,7 +338,54 @@ function notificationWidget($) {
         if (elem.hasClass('unread')) {
             markNotificationRead(elem.get(0));
         }
-        window.location = elem.data('location');
+
+        var container = document.querySelector('#container');
```
nemesifier, line 342: Why are we mixing vanilla JS with jQuery? If the code uses jQuery, we should be consistent, we can't start doing the same operations in different ways because it makes maintenance more costly.

openwisp_notifications/base/models.py
```diff
@@ -105,6 +105,10 @@ def target_url(self):
     def message(self):
         return self.get_message()
 
+    @cached_property
+    def rendered_description(self):
```
nemesifier, line 109: This looks good.

openwisp_notifications/api/serializers.py
```diff
@@ -59,6 +59,7 @@ class Meta(NotificationSerializer.Meta):
         fields = [
             'id',
             'message',
+            'rendered_description',
```
nemesifier, line 62: Here I am not convinced.

The API doesn't return any `description` property because we didn't use it until today, so I think it would be cleaner to define a new field called `description` which calls the `rendered_decscription` on the model instance (consult the Django REST Framework docs to see how to do this).

We'll only use the rendered description, it's highly unlikely we'll ever have to return the raw description field in the API, so let's simplify and hide this implementation details in the name of the field.

nemesifier (changes requested):

openwisp_notifications/static/openwisp-notifications/js/notifications.js
Dhanus3133: Updated

openwisp_notifications/api/serializers.py
Dhanus3133: Fixed it too

openwisp_notifications/static/openwisp-notifications/js/notifications.js
Dhanus3133: This allows the notification dialog or widget to close whenever the ESC is pressed. I would like to know if there is any other way to reduce it.

openwisp_notifications/static/openwisp-notifications/js/notifications.js
```diff
@@ -324,7 +348,41 @@ function notificationWidget($) {
...
+            document.querySelector('.ow-message-title').innerHTML = convertMessageWithRelativeURL(notification.message);
+            document.querySelector('.ow-message-description').innerHTML = notification.description;
+            $('.ow-dialog-overlay').removeClass('ow-hide');
+            if (notification.target_url) {
+                var target_url = new URL(notification.target_url, window.location.href);
+                $(document).on('click', '.ow-message-target-redirect', function () {
+                    window.location = target_url.pathname;
+                });
+                $('.ow-message-target-redirect').removeClass('ow-hide');
+            }
```
pandafy, line 372: I have this notification in my local environment 

```json
{
    "count": 1,
    "next": null,
    "previous": null,
    "results": [
        {
            "id": "f012ed5e-673b-4158-9cf0-d7292b72d10e",
            "message": "<p>Something alarming happened!</p>",
            "description": "<p>Test.</p>",
            "unread": false,
            "target_url": "#",
            "email_subject": "[external.openwisp.test] Generic Notification Subject",
            "timestamp": "2024-05-24T20:12:41.217258+02:00",
            "level": "info"
        }
    ]
}
```

As you can see, there's not a valid target_url. I don't think it makes sense to reload the page after clicking on "Open" when the target_url is set to `#`.

openwisp_notifications/tests/test_widget.py
pandafy: I think we can use that to add an additional CSS class to the notification element. If that class is present, the the logic for marking notification as read is skipped and it opens the notification dialog directly. We can also use a data attribute for this.

pandafy (changes requested):
Great work @Dhanus3133! I did an extensive code review where I also tried the code suggestions locally and tried to break the logic. For the most part, you have done a phenomenal job. 

I have left some suggestions below to improve the code implementation.

openwisp_notifications/tests/test_notifications.py
```diff
@@ -60,7 +60,6 @@ def setUp(self):
         self.notification_options = dict(
             sender=self.admin,
             description='Test Notification',
-            level='info',
```
pandafy, line 63: I went ahead and fixed the reason for failing test. This would allow us to close this PR quickly. 

In previous commits, we changed the logic of the `notify` handler to give preference to the `level` kwarg passed to the notify signal. 

@Dhanus3133 we need to add a test case in test_models.py which verifies that the preference of level is given to the `level` kwargs over the `level` defined in notification type.

openwisp_notifications/templates/admin/base_site.html
```diff
@@ -34,6 +34,16 @@
...
+    <div class="ow-dialog-overlay ow-overlay-inner ow-hide">
+        <div class="ow-dialog-notification">
+            <span class="ow-dialog-close ow-dialog-close-x">&times;</span>
+            <div class="ow-notification-level-wrapper ow-dialog-notification-level-wrapper"></div>
+            <h2 class="ow-message-title"></h2>
+            <div class="ow-message-description"></div>
+            <button class="ow-message-target-redirect ow-hide button default">{% trans 'Open' %}</button>
+            <button class="ow-dialog-close button default">{% trans 'Close' %}</button>
+        </div>
+    </div>
```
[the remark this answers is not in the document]
Dhanus3133: Seems yeah! Working fine.

openwisp_notifications/tests/test_widget.py
```diff
@@ -54,3 +53,28 @@ def test_notification_relative_link(self):
...
         )
+
+    def test_notification_dialog(self):
+        self.login()
+        self.notification_options.update(
+            {'message': 'Test Message', 'description': 'Test Description'}
+        )
+        notification = self._create_notification().pop()[1][0]
+        self.web_driver.find_element(By.ID, 'openwisp_notifications').click()
+        time.sleep(4)
```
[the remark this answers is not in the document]
Dhanus3133: Sorry, it was for debugging! Will remove it.

openwisp_notifications/static/openwisp-notifications/js/notifications.js
```diff
@@ -59,7 +61,9 @@ function initNotificationDropDown($) {
         // Check if the clicked area is dropDown / notification-btn or not
         if (
             $('.ow-notification-dropdown').has(e.target).length === 0 &&
-            !$(e.target).is($('.ow-notifications'))
+            !$(e.target).is($('.ow-notifications')) &&
+            !$(e.target).is($('.ow-dialog-close')) &&
+            !isNotificationDialogOpen
```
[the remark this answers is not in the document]
Dhanus3133: Didn't know these were available :)

openwisp_notifications/static/openwisp-notifications/js/notifications.js
```diff
@@ -69,24 +73,59 @@ function initNotificationDropDown($) {
     $(document).focusin(function(e){
         // Hide notification widget if focus is shifted to an element outside it
         e.stopPropagation();
-        if ($('.ow-notification-dropdown').has(e.target).length === 0){
+        if (
+            $('.ow-notification-dropdown').has(e.target).length === 0 &&
+            !isNotificationDialogOpen
+        ) {
             // Don't hide if focus changes to notification bell icon
             if (e.target != $('#openwisp_notifications').get(0)) {
```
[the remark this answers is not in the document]
Dhanus3133: This change makes the checks reusable, but it might add unnecessary complexity and confusion for now ig. If you need to then I will do it.

openwisp_notifications/static/openwisp-notifications/js/notifications.js
```diff
@@ -208,32 +247,17 @@ function notificationWidget($) {
...
-                        </div>
-                        <div class="ow-notification-date">${datetime}</div>
+                            <div class="ow-notification-level-wrapper">
+                                <div class="ow-notify-${elem.level} icon"></div>
+                                <div class="ow-notification-level-text">${elem.level}</div>
+                            </div>
+                            <div class="ow-notification-date">${datetime}</div>
                         </div>
-                    ${convertMessageWithRelativeURL(elem.message)}
+                    ${elem.description ? elem.message.replace(/<a [^>]*>([^<]*)<\/a>/g, '$1') : convertMessageWithRelativeURL(elem.message)}
```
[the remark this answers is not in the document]
Dhanus3133: Actually, in the call, we discussed on the generic message type we don't want to allow users to click on the `target_url` directly something like this on the `Org 2`
image
So, with this regex, we eliminate this ability. The users view the description and then proceed to the `target_url` via the dialog. (Will add comments there)

openwisp_notifications/static/openwisp-notifications/js/notifications.js
Dhanus3133: I will just make it disabled for `#` target_url

openwisp_notifications/tests/test_notifications.py
Dhanus3133: I believe the file is `test_notifications.py`. Updated with the changes there.

openwisp_notifications/tests/test_widget.py
Dhanus3133: I have tried a different approach to solving it, but is it okay to let it be `unread` even if the user has read that message?

openwisp_notifications/static/openwisp-notifications/js/notifications.js
```diff
@@ -69,24 +74,51 @@ function initNotificationDropDown($) {
...
         // Hide notification widget on "Escape" key
-        if (e.keyCode == 27){
+        if ($('.ow-dialog-overlay').is(':visible')) {
+            $('.ow-dialog-overlay').addClass('ow-hide');
+            $('.ow-message-target-redirect').addClass('ow-hide');
+        } else {
             $('.ow-notification-dropdown').addClass('ow-hide');
+        }
+        if ($(e.target).is($('.ow-notification-dropdown'))) {
             $('#openwisp_notifications').focus();
```
pandafy, line 102: Thinking out loud here: 

Shouldn't this be

```suggestion
            $('.ow-notification-dropdown').addClass('ow-hide');
            $('#openwisp_notifications').focus();
```

Why do we need to check `if ($(e.target).is($('.ow-notification-dropdown')))`

openwisp_notifications/tests/test_widget.py
pandafy: > I have tried a different approach to solving it, but is it okay to let it be `unread` even if the user has read that message?

No, it is not.

nemesifier (changes requested):
Great progress @Dhanus3133 @pandafy! :clap: 

I did some changes to the CSS. Please pull this change from openwisp-utils in your env: https://github.com/openwisp/openwisp-utils/commit/1a50adb03010f99b65a56bbb8dd5d47e0bfee335.

Here's the summary of my changes:

- As wrote previously, let's avoid to define colors as much as possible, colors should be defined generically in reusable classes as much as possible to avoid increasing maintenance overhead when we need to perform UX/theme/color  improvements.
- When adding new CSS rules, I wouldn't add them at the top of the file unless we're adding some kind of main style which is used in different places. Since we're adding a style which is specific to one element of the UI, let's add this towards the bottom of the file, preceded by a comment which summarizes what the style rules are for.
- I removed `.ow-dialog-overlay` because there's already a generic `.ow-overlay` style defined in the openwisp theme.
- I changed the open button to use a different color.
- I adjusted some spacing details

Here how it looks: 

Screenshot from 2024-05-29 18-30-50

Feel free to review my changes and let me know if there's any issue with them (feel free to fix).

I am pretty satisfied! But there's one outstanding issue I can see from my testing.

## Notification toast not opening dialog

I did already report this issue but while reviewing I think the screenshot I had attached was wrong.

Here's a GIF which should explain the issue better.

toast-dialog-not-working

I expect the generic notification dialog to open after clicking on the notification toast message, but instead it redirects to the target URL (ignore the 404 response, which is not relevant to your changes but it's just a dev env configuration issue).

Please address this outstanding issue so we can finally merge this great improvement.

Dhanus3133:
LGTM :rocket: ! @nemesifier

openwisp_notifications/tests/test_notifications.py
pandafy, line 1: Can we also add a test with the `generic_message` notification type?

Dhanus3133: I have added it.

nemesifier (changes requested):
The README update was missing. I added it, please have a look at it and use it as a reference for future modifications.

While writing the README I tried to come up with a simple example that doesn't specify the target object and I noticed the "Open" button is shown anyway, which is not right.

Please make sure that if the target is not specified, the "Open" button does not appear.

That way we can use the notification module to communicate any message to users, even if there isn't any object related to the message we're sending.

openwisp_notifications/base/models.py
```diff
@@ -105,36 +105,74 @@ def target_url(self):
...
+        try:
+            formatted_output = string.format(notification=self, **data)
+        except AttributeError as exception:
+            self._invalid_notification(
+                self.pk,
+                exception,
+                'Error encountered in rendering notification message',
+            )
+        self._delete_extra_attributes()
+        return formatted_output
```
pandafy, line 150: This is required because we handle the target URL differently when we send the email. 

The email includes a ridrect URL, which flags the notification as read and then redirects the user to the target object.

openwisp_notifications/base/models.py
```diff
@@ -105,36 +105,74 @@ def target_url(self):
...
+        if not self.description:
+            return
+        md_description = self._get_formatted_string(self.description)
+        return mark_safe(markdown(md_description))
+
     @property
     def email_message(self):
         return self.get_message(email_message=True)
 
+    def _set_extra_attributes(self, email_message=False):
```
pandafy, line 119: I thought of performing these operations in the `__init__`, but the target_link is dynamic based on whether we are using the message in the email.

openwisp_notifications/tests/test_notifications.py
```diff
@@ -323,6 +322,35 @@ def test_default_notification_type(self):
...
 
+    def test_generic_notification_type(self):
+        self.notification_options.pop('verb')
+        self.notification_options.update(
+            {
+                'message': '[{notification.actor}]({notification.actor_link})',
+                'type': 'generic_message',
+                'description': '[{notification.actor}]({notification.actor_link})',
+            }
+        )
```
pandafy, line 333: I improved this test to verity markdown rendering of message and description works.

openwisp_notifications/static/openwisp-notifications/js/notifications.js
```diff
@@ -387,6 +387,7 @@ function notificationHandler($, elem) {
 
     if (notification.target_url && notification.target_url !== '#') {
         targetUrl = new URL(notification.target_url).pathname;
+        $('.ow-message-target-redirect').removeClass('ow-hide');
```
nemesifier, line 390: Can you please add one new selenium test method only for this specific detail, which is quite important?

nemesifier (changes requested):
We are working with @pandafy to improve the changes he is making to allow him to accomplish what he needs to do in https://github.com/openwisp/openwisp-controller/pull/865.

@Dhanus3133 please see my last request below, with that I think we won't need your help anymore on this.

openwisp_notifications/static/openwisp-notifications/js/notifications.js
```diff
@@ -69,30 +75,54 @@ function initNotificationDropDown($) {
...
             // Don't hide if focus changes to notification bell icon
             if (e.target != $('#openwisp_notifications').get(0)) {
                 $('.ow-notification-dropdown').addClass('ow-hide');
             }
         }
     });
 
-    $('.ow-notification-dropdown').on('keyup', '*', function(e){
+    $(document).on('keyup', '*', function(e){
         e.stopPropagation();
```
pandafy, line 91: We should avoid doing such practices because it would break functionalities added by other modules. 

This is the reason for https://github.com/openwisp/openwisp-controller/issues/868 bug. 

I apologise that this exist!!!

nemesifier: I think we find a way to avoid to bind to the document like that, it should be possible to specify to which element to bind, or filter the elements instead of using `*`, probably also removing `e.stopPropagation();` may be a good idea, depending if we bind this event to an HTML container shared with the rest of the UI or with some other HTML container which is specific to the notifications module.

@Dhanus3133 please let me know if the suggestion above is clear.

Dhanus3133: Updated, let me know if that requires some changes.

nemesifier: It works, thanks!

openwisp_notifications/static/openwisp-notifications/js/notifications.js
```diff
@@ -69,30 +75,53 @@ function initNotificationDropDown($) {
     $(document).focusin(function(e){
```
nemesifier, line 75: @pandafy do you think we could avoid binding on `document` here too? That `stopPropagation()` below scares me. We could incur in the same problem we had for the keyup event in other modules if we need to use the same JS event. The problem will be that to find this will probably take hours because by that time we'll have forgotten.
We can open a new issue for this.

openwisp_notifications/static/openwisp-notifications/js/notifications.js
```diff
@@ -208,32 +238,18 @@ function notificationWidget($) {
...
-                        </div>
-                        <div class="ow-notification-date">${datetime}</div>
+                            <div class="ow-notification-level-wrapper">
+                                <div class="ow-notify-${elem.level} icon"></div>
+                                <div class="ow-notification-level-text">${elem.level}</div>
+                            </div>
+                            <div class="ow-notification-date">${datetime}</div>
                         </div>
-                    ${convertMessageWithRelativeURL(elem.message)}
+                        ${elem.description ? elem.message.replace(/<a [^>]*>([^<]*)<\/a>/g, '$1') : convertMessageWithRelativeURL(elem.message)}
```
nemesifier, line 252: @Dhanus3133 I find this line a bit cryptic.
Can we move the computation of this before the return statement using an extended if and making sure the comment about removing hyperlinks is right above the line doing the `replace`?

openwisp_notifications/static/openwisp-notifications/js/notifications.js
```diff
@@ -56,10 +57,15 @@ function initNotificationDropDown($) {
 
     $(document).click(function (e) {
         e.stopPropagation();
```
nemesifier, line 59: @pandafy I see a similar thing here :fearful:

nemesifier (commented):
See my comments below @pandafy @Dhanus3133.

openwisp_notifications/static/openwisp-notifications/js/notifications.js
Dhanus3133: Updated.

openwisp_notifications/static/openwisp-notifications/js/notifications.js
```diff
@@ -238,7 +238,14 @@ function notificationWidget($) {
         }
         klass = notificationReadStatus.get(elem.id);
 
-        // Remove hyperlinks from notification messages if description is present
+        // If description is present, remove hyperlinks from the message
```
nemesifier, line 241: The comment shouldn't be here but above the line which calls `replace`.
The part which says "if description is present" is redundant, the code is self explanatory and we don't need to add a comment that explains the if, we assume that anybody trying to read this code understands `if` statements.

openwisp_notifications/static/openwisp-notifications/js/notifications.js
```diff
@@ -238,7 +238,14 @@ function notificationWidget($) {
...
+        let message;
+        if (elem.description) {
+            message = elem.message.replace(/<a [^>]*>([^<]*)<\/a>/g, '$1');
```
nemesifier, line 244: The comment should be above this line, I would write something like `Remove hyperlinks from generic notifications to enforce the opening of the message dialog`.
This summarizes the work we've been doing and will help us remember why we did this even 2-3 years from now.
It's also useful for new developers who are studying the codebase.

nemesifier (changes requested):
Please see my comments below. I am insisting with these code comment details because it's important. If we add complexity without commenting the code properly we make maintenance of the system harder.

nemesifier (approved):
Great team work! @Dhanus3133 @pandafy :clap: :+1:  :tada:

PR #276, opened 2024-06-04
[feature] Batch email notifications

### LINKED ISSUE #132: [feature] Batch email notifications to prevent email flooding
Whenever there are infrastructure issues that cause lots of alerts, email boxes start to refuse receiving alert emails, which raise errors but is also bad because the email address sending the alerts could be flagged as spam.

The solution for this problem is to add the possibility to batch emails in a way that if there are more than X emails to the same user in less than X minutes, the subsequent emails get batched so they don't get sent singularly but rather summarized in a single email sent after the situation cools off (say X minutes after the user is not receiving emails).

A default configuration could be as follows:

- if more than 1 email has been sent to a specific user in the last 30 minutes
- subsequent emails to the same user get squashed in a summary, sending is paused for 30 minutes
- if another email is sent to the user, it will also be added to the summary which is on hold for sending and the timer will be reset (eg: 30 more minutes), this repeats until the timer expires and the summary is finally sent 

It's just an idea for now, we can improve it further.

A good example of an open source application which does this is sentry:

Screenshot from 2023-02-07 16-09-16

Dhanus3133:
Implements and closes #132.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1014,7 +1014,7 @@ The default configuration is as follows:
     }
 
 ``OPENWISP_NOTIFICATIONS_EMAIL_BATCH_INTERVAL``
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 +---------+-----------------------------------+
 | type    | ``int``                           |
@@ -1025,6 +1025,17 @@ The default configuration is as follows:
 This setting defines the interval at which the email notifications are sent in batches to users within the specified interval.
 
 If you want to send email notifications immediately, then set it to ``0``.
+
+``OPENWISP_NOTIFICATIONS_EMAIL_BATCH_DISPLAY_LIMIT``
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
++---------+-----------------------------------+
+| type    | ``int``                           |
++---------+-----------------------------------+
+| default | ``15``                            |
++---------+-----------------------------------+
+
+This setting defines the number of email notifications to be displayed in a batched email.
 
 Exceptions
 ----------
--- a/README.rst
+++ b/README.rst
@@ -1013,7 +1013,7 @@ The default configuration is as follows:
         'max_allowed_backoff': 15,
     }
 
-``EMAIL_BATCH_INTERVAL``
+``OPENWISP_NOTIFICATIONS_EMAIL_BATCH_INTERVAL``
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 +---------+-----------------------------------+
--- a/README.rst
+++ b/README.rst
@@ -1013,7 +1013,7 @@ The default configuration is as follows:
         'max_allowed_backoff': 15,
     }
 
-``OPENWISP_NOTIFICATIONS_EMAIL_BATCH_INTERVAL``
+``EMAIL_BATCH_INTERVAL``
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 +---------+-----------------------------------+
--- a/README.rst
+++ b/README.rst
@@ -956,6 +956,19 @@ The default configuration is as follows:
         'max_allowed_backoff': 15,
     }
 
+``OPENWISP_NOTIFICATIONS_EMAIL_BATCH_INTERVAL``
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
++---------+-----------------------------------+
+| type    | ``int``                           |
++---------+-----------------------------------+
+| default | ``1800`` `(30 mins, in seconds)`  |
++---------+-----------------------------------+
+
+This setting defines the interval at which the email notifications are sent in batches to users within the specified interval.
+
+If you want to send email notifications immediately, then set it to ``0``.
+
 Exceptions
 ----------
 
--- a/openwisp_notifications/handlers.py
+++ b/openwisp_notifications/handlers.py
@@ -25,7 +25,6 @@ from openwisp_notifications.websockets import handlers as ws_handlers
 logger = logging.getLogger(__name__)
 
 EXTRA_DATA = app_settings.get_config()['USE_JSONFIELD']
-EMAIL_BATCH_INTERVAL = app_settings.EMAIL_BATCH_INTERVAL
 
 User = get_user_model()
 
@@ -205,6 +204,7 @@ def send_email_notification(sender, instance, created, **kwargs):
             'email_id': instance.recipient.email,
         },
     )
+    EMAIL_BATCH_INTERVAL = app_settings.EMAIL_BATCH_INTERVAL
 
     if cache_data['last_email_sent_time'] and EMAIL_BATCH_INTERVAL > 0:
         # Case 1: Batch email sending logic
--- a/openwisp_notifications/handlers.py
+++ b/openwisp_notifications/handlers.py
@@ -10,7 +10,6 @@ from django.db.models.query import QuerySet
 from django.db.models.signals import post_delete, post_save
 from django.dispatch import receiver
 from django.utils import timezone
-from django.utils.translation import gettext as _
 
 from openwisp_notifications import settings as app_settings
 from openwisp_notifications import tasks
@@ -20,13 +19,13 @@ from openwisp_notifications.types import (
     NOTIFICATION_ASSOCIATED_MODELS,
     get_notification_configuration,
 )
+from openwisp_notifications.utils import send_notification_email
 from openwisp_notifications.websockets import handlers as ws_handlers
-from openwisp_utils.admin_theme.email import send_email
 
 logger = logging.getLogger(__name__)
 
 EXTRA_DATA = app_settings.get_config()['USE_JSONFIELD']
-EMAIL_BATCH_INTERVAL = app_settings.OPENWISP_NOTIFICATIONS_EMAIL_BATCH_INTERVAL
+EMAIL_BATCH_INTERVAL = app_settings.EMAIL_BATCH_INTERVAL
 
 User = get_user_model()
 
@@ -165,36 +164,6 @@ def notify_handler(**kwargs):
     return notification_list
 
 
-def send_single_email(instance):
-    try:
-        subject = instance.email_subject
-    except NotificationRenderException:
-        # Do not send email if notification is malformed.
-        return
-    url = instance.data.get('url', '') if instance.data else None
-    description = instance.message
-    if url:
-        target_url = url
-    elif instance.target:
-        target_url = instance.redirect_view_url
-    else:
-        target_url = None
-    if target_url:
-        description += _('\n\nFor more information see %(target_url)s.') % {
-            'target_url': target_url
-        }
-    send_email(
-        subject,
-        description,
-        instance.message,
-        recipients=[instance.recipient.email],
-        extra_context={
-            'call_to_action_url': target_url,
-            'call_to_action_text': _('Find out more'),
-        },
-    )
-
-
 @receiver(post_save, sender=Notification, dispatch_uid='send_email_notification')
 def send_email_notification(sender, instance, created, **kwargs):
     # Abort if a new notification is not created
@@ -223,32 +192,37 @@ def send_email_notification(sender, instance, created, **kwargs):
     if not (email_preference and instance.recipient.email and email_verified):
         return
 
-    recipient_id = instance.recipient.id
-    cache_key = f'email_batch_{recipient_id}'
-    cache_key_pks = f'{instance.recipient.email}_pks'
+    recipient_email = instance.recipient.email
+    cache_key = f'email_batch_{recipient_email}'
+    cache_key_pks = f'{recipient_email}_batch_pks'
     cache_data = cache.get(
         cache_key, {'last_email_sent_time': None, 'batch_scheduled': False}
     )
 
     if cache_data['last_email_sent_time'] and EMAIL_BATCH_INTERVAL > 0:
+        # Case 1: Batch email sending logic
         if not cache_data['batch_scheduled']:
-            # Schedule batch email notification task
+            # Schedule batch email notification task if not already scheduled
             tasks.batch_email_notification.apply_async(
                 (instance.recipient.email,), countdown=EMAIL_BATCH_INTERVAL
             )
+            # Mark batch as scheduled to prevent duplicate scheduling
             cache_data['batch_scheduled'] = True
-            cache.set(cache_key_pks, [instance.id], timeout=EMAIL_BATCH_INTERVAL * 2)
+            cache.set(cache_key_pks, [instance.id])  # Initialize list of IDs for batch
             cache.set(cache_key, cache_data, timeout=EMAIL_BATCH_INTERVAL)
         else:
+            # Add current instance ID to the list of IDs for batch
             ids = cache.get(cache_key_pks, [])
             ids.append(instance.id)
-            cache.set(cache_key_pks, ids, timeout=EMAIL_BATCH_INTERVAL * 2)
+            cache.set(cache_key_pks, ids)
         return
 
+    # Case 2: Single email sending logic
+    # Update the last email sent time and cache the data
     cache_data['last_email_sent_time'] = timezone.now()
     cache.set(cache_key, cache_data, timeout=EMAIL_BATCH_INTERVAL)
 
-    send_single_email(instance)
+    send_notification_email(instance)
 
     # flag as emailed
     instance.emailed = True
--- a/openwisp_notifications/handlers.py
+++ b/openwisp_notifications/handlers.py
@@ -162,6 +162,36 @@ def notify_handler(**kwargs):
     return notification_list
 
 
+def send_single_email(instance):
+    try:
+        subject = instance.email_subject
+    except NotificationRenderException:
+        # Do not send email if notification is malformed.
+        return
+    url = instance.data.get('url', '') if instance.data else None
+    description = instance.message
+    if url:
+        target_url = url
+    elif instance.target:
+        target_url = instance.redirect_view_url
+    else:
+        target_url = None
+    if target_url:
+        description += _('\n\nFor more information see %(target_url)s.') % {
+            'target_url': target_url
+        }
+    send_email(
+        subject,
+        description,
+        instance.message,
+        recipients=[instance.recipient.email],
+        extra_context={
+            'call_to_action_url': target_url,
+            'call_to_action_text': _('Find out more'),
+        },
+    )
+
+
 @receiver(post_save, sender=Notification, dispatch_uid='send_email_notification')
 def send_email_notification(sender, instance, created, **kwargs):
     # Abort if a new notification is not created
@@ -190,23 +220,7 @@ def send_email_notification(sender, instance, created, **kwargs):
     if not (email_preference and instance.recipient.email and email_verified):
         return
 
-    try:
-        subject = instance.email_subject
-    except NotificationRenderException:
-        # Do not send email if notification is malformed.
-        return
-    url = instance.data.get('url', '') if instance.data else None
-    description = instance.message
-    if url:
-        target_url = url
-    elif instance.target:
-        target_url = instance.redirect_view_url
-    else:
-        target_url = None
-    if target_url:
-        description += _('\n\nFor more information see %(target_url)s.') % {
-            'target_url': target_url
-        }
+    send_single_email(instance)
 
     recipient_id = instance.recipient.id
     cache_key = f'email_batch_{recipient_id}'
@@ -232,16 +246,9 @@ def send_email_notification(sender, instance, created, **kwargs):
 
     cache_data['last_email_sent_time'] = timezone.now()
     cache.set(cache_key, cache_data, timeout=EMAIL_BATCH_INTERVAL)
-    send_email(
-        subject,
-        description,
-        instance.message,
-        recipients=[instance.recipient.email],
-        extra_context={
-            'call_to_action_url': target_url,
-            'call_to_action_text': _('Find out more'),
-        },
-    )
+
+    send_single_email(instance)
+
     # flag as emailed
     instance.emailed = True
     # bulk_update is used to prevent emitting post_save signal
--- a/openwisp_notifications/handlers.py
+++ b/openwisp_notifications/handlers.py
@@ -25,7 +25,7 @@ from openwisp_notifications.websockets import handlers as ws_handlers
 logger = logging.getLogger(__name__)
 
 EXTRA_DATA = app_settings.get_config()['USE_JSONFIELD']
-EMAIL_BATCH_INTERVAL = app_settings.OPENWISP_NOTIFICATIONS_EMAIL_BATCH_INTERVAL
+EMAIL_BATCH_INTERVAL = app_settings.EMAIL_BATCH_INTERVAL
 
 User = get_user_model()
 
@@ -201,6 +201,7 @@ def send_email_notification(sender, instance, created, **kwargs):
             'last_email_sent_time': None,
             'batch_scheduled': False,
             'pks': [],
+            'start_time': None,
             'email_id': instance.recipient.email,
         },
     )
@@ -209,12 +210,13 @@ def send_email_notification(sender, instance, created, **kwargs):
         # Case 1: Batch email sending logic
         if not cache_data['batch_scheduled']:
             # Schedule batch email notification task if not already scheduled
-            tasks.batch_email_notification.apply_async(
+            tasks.send_batched_email_notifications.apply_async(
                 (instance.recipient.id,), countdown=EMAIL_BATCH_INTERVAL
             )
             # Mark batch as scheduled to prevent duplicate scheduling
             cache_data['batch_scheduled'] = True
             cache_data['pks'] = [instance.id]
+            cache_data['start_time'] = timezone.now()
             cache.set(cache_key, cache_data)
         else:
             # Add current instance ID to the list of IDs for batch
@@ -224,8 +226,9 @@ def send_email_notification(sender, instance, created, **kwargs):
 
     # Case 2: Single email sending logic
     # Update the last email sent time and cache the data
-    cache_data['last_email_sent_time'] = timezone.now()
-    cache.set(cache_key, cache_data, timeout=EMAIL_BATCH_INTERVAL)
+    if EMAIL_BATCH_INTERVAL > 0:
+        cache_data['last_email_sent_time'] = timezone.now()
+        cache.set(cache_key, cache_data, timeout=EMAIL_BATCH_INTERVAL)
 
     send_notification_email(instance)
 
--- a/openwisp_notifications/handlers.py
+++ b/openwisp_notifications/handlers.py
@@ -1,7 +1,6 @@
 import logging
 
 from celery.exceptions import OperationalError
-from celery.result import AsyncResult
 from django.contrib.auth import get_user_model
 from django.contrib.contenttypes.models import ContentType
 from django.core.cache import cache
@@ -27,7 +26,7 @@ from openwisp_utils.admin_theme.email import send_email
 logger = logging.getLogger(__name__)
 
 EXTRA_DATA = app_settings.get_config()['USE_JSONFIELD']
-EMAIL_BATCH_INTERVAL = app_settings.get_config()['EMAIL_BATCH_INTERVAL']
+EMAIL_BATCH_INTERVAL = app_settings.OPENWISP_NOTIFICATIONS_EMAIL_BATCH_INTERVAL
 
 User = get_user_model()
 
@@ -163,8 +162,12 @@ def notify_handler(**kwargs):
     return notification_list
 
 
-def can_send_email(instance):
-    # Get email preference of the user for this type of notification
+@receiver(post_save, sender=Notification, dispatch_uid='send_email_notification')
+def send_email_notification(sender, instance, created, **kwargs):
+    # Abort if a new notification is not created
+    if not created:
+        return
+    # Get email preference of user for this type of notification.
     target_org = getattr(getattr(instance, 'target', None), 'organization_id', None)
     if instance.type and target_org:
         try:
@@ -172,28 +175,26 @@ def can_send_email(instance):
                 organization=target_org, type=instance.type
             )
         except NotificationSetting.DoesNotExist:
-            return False
+            return
         email_preference = notification_setting.email_notification
     else:
-        # Send email anyway if notification type or target_org is absent
+        # We can not check email preference if notification type is absent,
+        # or if target_org is not present
+        # therefore send email anyway.
         email_preference = True
 
-    # Check if email is verified
     email_verified = instance.recipient.emailaddress_set.filter(
         verified=True, email=instance.recipient.email
     ).exists()
 
-    # Verify email preference and email verification status
     if not (email_preference and instance.recipient.email and email_verified):
-        return False
+        return
 
-    # Verify the email subject
     try:
         subject = instance.email_subject
     except NotificationRenderException:
-        return False
-
-    # Get the URL and description
+        # Do not send email if notification is malformed.
+        return
     url = instance.data.get('url', '') if instance.data else None
     description = instance.message
     if url:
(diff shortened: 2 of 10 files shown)


openwisp_notifications/settings.py
```diff
@@ -37,6 +37,12 @@
     'openwisp-notifications/audio/notification_bell.mp3',
 )
 
+EMAIL_BATCH_INTERVAL = getattr(
+    settings,
+    'EMAIL_BATCH_INTERVAL',
```
pandafy, line 42: This should be `OPENWISP_NOTIFICATIONS_EMAIL_BATCH_INTERVAL`

openwisp_notifications/handlers.py
```diff
@@ -26,6 +27,7 @@
 logger = logging.getLogger(__name__)
 
 EXTRA_DATA = app_settings.get_config()['USE_JSONFIELD']
+EMAIL_BATCH_INTERVAL = app_settings.get_config()['EMAIL_BATCH_INTERVAL']
```
pandafy, line 30: The `app_settings.get_config()` returns the configuration for the Notification model. It is a django-notiifcations thing. 

Since you are defining the setting globally in openwisp_notifications/settings.py. You can access the setting like an attribute. 

```suggestion
EMAIL_BATCH_INTERVAL = app_settings.EMAIL_BATCH_INTERVAL
```

openwisp_notifications/tasks.py
```diff
@@ -3,14 +3,21 @@
...
 
 from openwisp_notifications import types
+from openwisp_notifications import settings as app_settings
+from openwisp_notifications import handlers
 from openwisp_notifications.swapper import load_model, swapper_load_model
+from openwisp_notifications import settings as app_settings
+from openwisp_utils.admin_theme.email import send_email
 from openwisp_utils.tasks import OpenwispCeleryTask
 
+EMAIL_BATCH_INTERVAL = app_settings.get_config()['EMAIL_BATCH_INTERVAL']
```
pandafy, line 19:
```suggestion
EMAIL_BATCH_INTERVAL = app_settings.EMAIL_BATCH_INTERVAL
```

openwisp_notifications/tasks.py
```diff
@@ -3,14 +3,21 @@
...
+from django.core.cache import cache
 from django.db.models import Q
 from django.db.utils import OperationalError
 from django.utils import timezone
 
 from openwisp_notifications import types
+from openwisp_notifications import settings as app_settings
+from openwisp_notifications import handlers
 from openwisp_notifications.swapper import load_model, swapper_load_model
+from openwisp_notifications import settings as app_settings
```
pandafy, line 15: There is a duplicate import

openwisp_notifications/tasks.py
```diff
@@ -3,14 +3,21 @@
...
 from django.contrib.auth import get_user_model
 from django.contrib.contenttypes.models import ContentType
+from django.core.cache import cache
 from django.db.models import Q
 from django.db.utils import OperationalError
 from django.utils import timezone
 
 from openwisp_notifications import types
+from openwisp_notifications import settings as app_settings
+from openwisp_notifications import handlers
```
pandafy, line 13: Won't this trigger circular imports error?

openwisp_notifications/handlers.py
```diff
@@ -203,25 +203,61 @@ def send_email_notification(sender, instance, created, **kwargs):
...
+    return True, subject, description, target_url
+
+
+@receiver(post_save, sender=Notification, dispatch_uid='send_email_notification')
+def send_email_notification(sender, instance, created, **kwargs):
+    # Abort if a new notification is not created
+    if not created:
+        return
+
+    can_send, subject, description, target_url = can_send_email(instance)
```
pandafy, line 217: Why did we require to refactor this logic?

I am not convinced by the return signature of the `can_send_email` function. E.g. if the notification is malformed, `can_send_email` will only return `False` (1 return value). I believe the tuple unpacking would fail in this scenario.

openwisp_notifications/handlers.py
```diff
@@ -203,25 +203,61 @@ def send_email_notification(sender, instance, created, **kwargs):
...
+
+    if not can_send:
+        return
+
+    recipient_id = instance.recipient.id 
+    cache_key = f'email_batch_{recipient_id}'
+    cache_data = cache.get(cache_key, {'last_email_sent_time': None, 'batch_scheduled': False})
+    is_send_email = False
+
+    if cache_data['last_email_sent_time']:
```
pandafy, line 227: I think we could simplify these conditionals. 

If `cache.get(cache_key)` returns `None`, this means one of  the two things 
1. No email was sent to the user 
2.  The `EMAIL_BATCH_INTERVAL` has passed because you are setting a timeout to the cache in this line 
```
cache.set(cache_key, cache_data, timeout=EMAIL_BATCH_INTERVAL)
```

I think, we can handle both of these scenarios in the same condition.

What do you think @Dhanus3133? Will it cover all the scenarios?

openwisp_notifications/handlers.py
```diff
@@ -203,25 +203,61 @@ def send_email_notification(sender, instance, created, **kwargs):
...
+        if (timezone.now() - cache_data['last_email_sent_time']).seconds < EMAIL_BATCH_INTERVAL:
+            if not cache_data['batch_scheduled']:
```
pandafy, line 229: The fact that the `cache_data['last_email_sent_time']` exists should be enough to tell there's an scheduled task for the batch sending. 

We would need some comments here to explain all of this.

pandafy (changes requested):
@Dhanus3133 I have left some suggestions. I hope these solve your queries.

nemesifier (commented):
@Dhanus3133 please start writin a short README section that explains how this feature works and add the new setting.
PS: once we're done with these changes, please add a screenshot in the README.

@pandafy you know what would be handy?

- Allow to change the batching using an organization setting that allows user to change this per org via the admin
- Allow to specify whether the system should start batching right away instead of waiting 1 email, this could be disabled by default, and could also be available from the organization page

openwisp_notifications/handlers.py
```diff
@@ -26,6 +26,7 @@
 logger = logging.getLogger(__name__)
 
 EXTRA_DATA = app_settings.get_config()['USE_JSONFIELD']
+EMAIL_BATCH_INTERVAL = app_settings.OPENWISP_NOTIFICATIONS_EMAIL_BATCH_INTERVAL
```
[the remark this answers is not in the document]
Dhanus3133: Updated.

Dhanus3133:
In our weekly call, @nemesifier and @pandafy  suggested some improvements.

1. The date should be displayed below the title. 
2. Update the email title with the number of notifications and site name.
3. Some changes have to be made regarding the email template
    1. If there is only one notification in the batch, just use a single mail template.
    2. Can follow the current template if there are 2 - 5 notifications.
    3. If there are more than 5 emails, a new template can be used where only the level, message, and date can be displayed. Along with a ‘View all new notifications’ link to the admin site.

openwisp_notifications/handlers.py
```diff
@@ -164,6 +165,36 @@ def notify_handler(**kwargs):
     return notification_list
 
 
+def send_single_email(instance):
```
pandafy, line 168: Can we move this function to utils.py?  I would rename it to "send_notification_email" and add docstring to the function which explains that this function should be used when there's only one notification for the recipient.

openwisp_notifications/handlers.py
```diff
@@ -192,34 +223,32 @@ def send_email_notification(sender, instance, created, **kwargs):
...
+    )
+
+    if cache_data['last_email_sent_time'] and EMAIL_BATCH_INTERVAL > 0:
+        if not cache_data['batch_scheduled']:
+            # Schedule batch email notification task
+            tasks.batch_email_notification.apply_async(
+                (instance.recipient.email,), countdown=EMAIL_BATCH_INTERVAL
+            )
+            cache_data['batch_scheduled'] = True
+            cache.set(cache_key_pks, [instance.id], timeout=EMAIL_BATCH_INTERVAL * 2)
```
pandafy, line 240: We should not set a timeout for the cache here. It may be possible that the celery worker is overloaded. It could take more time than the set timeout for execution of the batch email sending task. We should only clear the cache key once the batch email processing is complete.

openwisp_notifications/handlers.py
```diff
@@ -192,34 +223,32 @@ def send_email_notification(sender, instance, created, **kwargs):
     if not (email_preference and instance.recipient.email and email_verified):
         return
 
-    try:
-        subject = instance.email_subject
-    except NotificationRenderException:
-        # Do not send email if notification is malformed.
+    recipient_id = instance.recipient.id
+    cache_key = f'email_batch_{recipient_id}'
+    cache_key_pks = f'{instance.recipient.email}_pks'
```
pandafy, line 228: Why did you use a separate key for storing PKs? Can we add another field to `cache_key`?

openwisp_notifications/handlers.py
```diff
@@ -192,34 +223,32 @@ def send_email_notification(sender, instance, created, **kwargs):
...
+    cache_data = cache.get(
+        cache_key, {'last_email_sent_time': None, 'batch_scheduled': False}
+    )
+
+    if cache_data['last_email_sent_time'] and EMAIL_BATCH_INTERVAL > 0:
```
pandafy, line 233: It would be helpful to add some comments here explaining why this logic exists and describing the two cases it handles.

openwisp_notifications/settings.py
```diff
@@ -37,6 +37,10 @@
     'openwisp-notifications/audio/notification_bell.mp3',
 )
 
+OPENWISP_NOTIFICATIONS_EMAIL_BATCH_INTERVAL = getattr(
+    settings, 'EMAIL_BATCH_INTERVAL', 30 * 60  # 30 minutes
```
pandafy, line 41: **Thinking out loud**

I think 30 minutes is a lot of time. It will affect the real-timeness of the email notifications. Most of the consecutive notifications would be delayed by this time. We should also point this out in the README. 


@nemesifier what do you think?

pandafy (changes requested):
This is looking good. I have done a quick code review and left some suggestion on improving the code.

I will do a function test tomorrow  and report if I find something to improve.

openwisp_notifications/tasks.py
```diff
@@ -202,3 +211,70 @@ def delete_ignore_object_notification(instance_id):
...
+            description = notification.message if notifications_count <= 5 else None
+            alerts.append(
+                {
+                    'level': notification.level,
+                    'message': notification.message,
+                    'description': description,
+                    'timestamp': notification.timestamp,
+                    'url': target_url,
+                }
+            )
```
pandafy, line 253: The only additional fields in this dictionary are the url and description fields.

You can set the url attribute on the notification object like this and use it in the template:

```python
    for notification in unsent_notifications:
            url = notification.data.get('url', '') if notification.data else None
            if url:
                notification.url = url
            elif notification.target:
                notification.url = notification.redirect_view_url
            else:
                notification.url = None
```

For the description, there's already a conditional in the template checking if alert.description. Instead of this, we can check the notifications_count in that conditional. However, I prefer to keep this logic in the Python code to avoid it becoming obscure in the future.

I recommend adding another context variable, show_notification_message (or a better name if you prefer), which is a boolean: show_notification_message = notifications_count <= 5. Please also add a comment explaining that the notification message is hidden in the email for brevity.

openwisp_notifications/tasks.py
```diff
@@ -202,3 +211,70 @@ def delete_ignore_object_notification(instance_id):
...
+
+        context = {
+            'alerts': alerts,
+            'notifications_count': notifications_count,
+            'site_name': current_site.name,
+            'site_domain': current_site.domain,
+        }
+        html_content = render_to_string('emails/batch_email.html', context)
+        plain_text_content = strip_tags(html_content)
```
pandafy, line 262: The plain text email looks like this 

```
Content-Type: multipart/alternative;
 boundary="===============8141022284195897303=="
MIME-Version: 1.0
Subject: Summary of 3 Notifications
From: webmaster@localhost
To: admin@admin.com
Date: Tue, 18 Jun 2024 07:30:02 -0000
Message-ID: <171869580268.41107.14558799017281335812@FRIDAY>

--===============8141022284195897303==
Content-Type: text/plain; charset="utf-8"
MIME-Version: 1.0
Content-Transfer-Encoding: 7bit



        .container-alerts {
            margin: 0 auto;
            background-color: #fff;
            padding: 20px;
            border: 1px solid #ddd;
            box-shadow: 0 0 10px rgba(0, 0, 0, 0.1);
        }
        .header {
            text-align: center;
            padding: 10px 0;
            border-bottom: 1px solid #ddd;
            margin-bottom: 20px;
        }
        .header h1 {
            margin: 0;
            font-size: 24px;
        }
        .header p {
            margin: 5px 0 0 0;
            font-size: 14px;
            color: #666;
        }
        .alert {
            border: 1px solid #e0e0e0;
            border-radius: 5px;
            margin-bottom: 10px;
            padding: 10px;
        }
        .alert.error {
            background-color: #ffefef;
        }
        .alert.info {
            background-color: #f0f0f0;
        }
        .alert.success {
            background-color: #e6f9e8;
        }
        .alert h2 {
            margin: 0 0 5px 0;
            font-size: 16px;
        }
        .alert h2 .title {
            display: inline-block;
            max-width: 80%;
            white-space: nowrap;
            overflow: hidden;
            text-overflow: ellipsis;
            vertical-align: middle;
        }
        .alert.error h2 {
            color: #d9534f;
        }
        .alert.info h2 {
            color: #333333;
        }
        .alert.success h2 {
            color: #1c8828;
        }
        .alert p {
            margin: 0;
            font-size: 14px;
            color: #666;
        }
        .alert .title p {
            display: inline;
            overflow: hidden;
            text-overflow: ellipsis;
        }
        .badge {
            display: inline-block;
            padding: 2px 8px;
            border-radius: 3px;
            font-size: 12px;
            font-weight: bold;
            text-transform: uppercase;
            margin-right: 5px;
            color: white;
        }
        .badge.error {
            background-color: #d9534f;
        }
        .badge.info {
            background-color: #333333;
        }
        .badge.success {
            background-color: #1c8828;
        }
        .alert a {
            text-decoration: none;
        }
        .alert.error a {
            color: #d9534f;
        }
        .alert.info a {
            color: #333333;
        }
        .alert.success a {
            color: #1c8828;
        }
        .alert a:hover {
            text-decoration: underline;
        }






            Summary of 3 Notifications
            From example.com




                INFO


                        Default notification with default verb and level info by default



            June 18, 2024, 9:29 a.m.

                Default notification with default verb and level info by default





                INFO


                        Default notification with default verb and level info by default



            June 18, 2024, 9:29 a.m.

                Default notification with default verb and level info by default





                INFO


                        Default notification with default verb and level info by default



            June 18, 2024, 9:29 a.m.

                Default notification with default verb and level info by default

```

It shall not contain the CSS block. The `strip_tags` is not removing them from the rendered string. We may need to add a dedicated template for the plain text email. 

Let's discuss this in today's weekly meeting before proceeding.

openwisp_notifications/utils.py
```diff
@@ -28,3 +32,36 @@ def normalize_unread_count(unread_count):
         return '99+'
     else:
         return unread_count
+
+
+def send_notification_email(instance):
```
pandafy, line 37:
```suggestion
def send_notification_email(notification):
```

openwisp_notifications/tests/test_notifications.py
```diff
@@ -944,6 +944,22 @@ def test_notification_for_unverified_email(self):
         # we don't send emails to unverified email addresses
         self.assertEqual(len(mail.outbox), 0)
 
+    @patch('openwisp_notifications.tasks.batch_email_notification.apply_async')
+    def test_batch_email_notification(self, mock_send_email):
```
pandafy, line 948: I would like to see some more test cases, testing different scenarios (sending batch email or single email). How the email template renders for 5+ notifications (description is absent) and 10 notifications (call to action button is present).

pandafy (changes requested):
### Immediate email after batch email 

If a notification is created just after the batch email is processed, then the email for this notification is sent immediately after the batch email. 

Is this a desired behaviour? Can you please add a test for this?

### Duplicate subject in email

The send_email function adds a subject in the email which servers as a heading for all the email communication.  

Screenshot from 2024-06-18 13-05-59


I think, we should keep using the subject provided by the `send_email` function for consistency.

pandafy (changes requested):
@Dhanus3133  the test is failing because the notification email contains direct URL to the target object while the test expects redirect url

```diff
-<div class="msg"><p>Default notification with default verb and level info by <a href="https://example.com/api/v1/notifications/notification/2632e308-0f2d-43a1-8617-7350e48b8fd3/redirect/">Tester Tester (test org)</a></p></div>
+<div class="msg"><p>Default notification with default verb and level info by <a href="https://example.com/admin/openwisp_users/organizationuser/dea5c3cc-8c2a-4021-94a7-9cc36885f9ff/change/">Tester Tester (test org)</a></p></div>

File: README.rst [hunk 19]
```diff
@@ -1013,6 +1013,19 @@ The default configuration is as follows:
         'max_allowed_backoff': 15,
     }
 
+``OPENWISP_NOTIFICATIONS_EMAIL_BATCH_INTERVAL``
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
nemesifier, line 1017: This seems longer than it should

openwisp_notifications/handlers.py
```diff
@@ -192,35 +192,43 @@ def send_email_notification(sender, instance, created, **kwargs):
...
+        else:
+            # Add current instance ID to the list of IDs for batch
+            cache_data['pks'].append(instance.id)
+            cache.set(cache_key, cache_data)
+        return
+
+    # Case 2: Single email sending logic
+    # Update the last email sent time and cache the data
+    cache_data['last_email_sent_time'] = timezone.now()
+    cache.set(cache_key, cache_data, timeout=EMAIL_BATCH_INTERVAL)
```
nemesifier, line 228: if `EMAIL_BATCH_INTERVAL` is zero, wouldn't this data be cached indefinitely? If that's the case, it's a problem.

openwisp_notifications/tasks.py
```diff
@@ -202,3 +210,67 @@ def delete_ignore_object_notification(instance_id):
     Deletes IgnoreObjectNotification object post it's expiration.
     """
     IgnoreObjectNotification.objects.filter(id=instance_id).delete()
+
+
+@shared_task(base=OpenwispCeleryTask)
+def batch_email_notification(instance_id):
```
nemesifier, line 216: I am not convinced of the name of this function here.

Does this function batch emails, or does it send batched emails? From the name it seems it does the former, but from the code it seems to me it does the latter. Please clarify.

If it's the latter, we should rename it to something like `send_batched_email_notifications` or something similar.

openwisp_notifications/settings.py
```diff
@@ -37,6 +37,10 @@
     'openwisp-notifications/audio/notification_bell.mp3',
 )
 
+OPENWISP_NOTIFICATIONS_EMAIL_BATCH_INTERVAL = getattr(
```
nemesifier, line 40: Let's shorten this, we usually do this in most modules, eg:

```suggestion
EMAIL_BATCH_INTERVAL = getattr(
```

openwisp_notifications/tasks.py
```diff
@@ -202,3 +210,67 @@ def delete_ignore_object_notification(instance_id):
...
+            'notifications': unsent_notifications,
+            'notifications_count': notifications_count,
+            'show_notification_description': show_notification_description,
+            'site_name': current_site.name,
+        }
+        html_content = render_to_string('emails/batch_email.html', context)
+        plain_text_content = render_to_string('emails/batch_email.txt', context)
+
+        extra_context = {}
+        if notifications_count > 5:
```
nemesifier, line 260: are we showing only 5 notifications? Let's show more and make this configurable with a setting, eg: `OPENWISP_NOTIFICATIONS_EMAIL_BATCH_DISPLAY_LIMIT` (add to README). I think we can default this to 15.

openwisp_notifications/tasks.py
```diff
@@ -202,3 +210,67 @@ def delete_ignore_object_notification(instance_id):
...
+            extra_context = {
+                'call_to_action_url': f"https://{current_site.domain}/admin",
+                'call_to_action_text': 'View all Notifications',
```
nemesifier, line 263: Always flag strings as translatable:
```suggestion
                'call_to_action_text': _('View all Notifications'),
```

openwisp_notifications/tasks.py
```diff
@@ -202,3 +210,67 @@ def delete_ignore_object_notification(instance_id):
...
+            }
+
+        send_email(
+            subject=f'Summary of {notifications_count} Notifications from {current_site.name}',
```
nemesifier, line 267: Same here.
Let's change the message to:

`[{current_site.name}] {notifications_count} new notifications since <UTC_DATE_AND_TIME>`, date and time showing like "July 1, 2024, 4:15 a.m. UTC".

openwisp_notifications/templates/emails/batch_email.txt
```diff
@@ -0,0 +1,15 @@
...
+{% for notification in notifications %}
+- {{ notification.message }}
+  URL: {{ notification.url }}
+  Timestamp: {{ notification.timestamp|date:"F j, Y, g:i a" }}
+  {% if show_notification_description %}
+      Description: {{ notification.rendered_description }}
+  {% endif %}
+{% endfor %}
+
+{% if notifications_count > 5 %}
```
nemesifier, line 12: Use the setting here instead of hardcoding 5

openwisp_notifications/templates/emails/batch_email.txt
```diff
@@ -0,0 +1,15 @@
+Summary of {{ notifications_count }} Notifications from {{ site_name }}
```
nemesifier, line 1: Flag all strings as translatable using `translate` and `blocktranslate`: https://docs.djangoproject.com/en/5.0/topics/i18n/translation/#internationalization-in-template-code

openwisp_notifications/utils.py
```diff
@@ -28,3 +32,36 @@ def normalize_unread_count(unread_count):
         return '99+'
     else:
         return unread_count
+
+
+def send_notification_email(notification):
+
```
nemesifier, line 38: can we remove this blank line or does `black` complain about it?

openwisp_notifications/tasks.py
```diff
@@ -202,3 +210,67 @@ def delete_ignore_object_notification(instance_id):
...
+    notifications_count = unsent_notifications.count()
+    current_site = Site.objects.get_current()
+    email_id = cache_data.get('email_id')
+
+    # Send individual email if there is only one notification
+    if notifications_count == 1:
+        notification = unsent_notifications.first()
+        send_notification_email(notification)
+    else:
+        # Show notification description upto 5 notifications
```
nemesifier, line 239:
```suggestion
        # Show the amount of notifications according to configured display limit
```

openwisp_notifications/tasks.py
```diff
@@ -202,3 +210,67 @@ def delete_ignore_object_notification(instance_id):
...
+        show_notification_description = notifications_count <= 5
```
nemesifier, line 240: use setting instead of hardcoding 5

openwisp_notifications/tests/test_notifications.py
```diff
@@ -944,6 +944,46 @@ def test_notification_for_unverified_email(self):
...
+
+        # Call the task
+        tasks.batch_email_notification(self.admin.id)
+
+        # Check if the rest of the notifications are sent in a batch
+        self.assertEqual(len(mail.outbox), 2)
+        self.assertIn('Summary of 10 Notifications', mail.outbox[1].subject)
+        self.assertIn('View all Notifications', mail.outbox[1].body)
+        self.assertNotIn('Test Notification', mail.outbox[1].body)
+
```
nemesifier, line 986: Can we add another test in which this feature is disabled and ensure that no batching happens in that case?
Using the `override_settings` from django or mocking the setting should work.

nemesifier (changes requested):
Great work @Dhanus3133 @pandafy, see my comments below.

README.rst
Dhanus3133: I assumed that it would be okay since we already use this setting [`OPENWISP_NOTIFICATIONS_POPULATE_PREFERENCES_ON_MIGRATE`](https://github.com/openwisp/openwisp-notifications/#openwisp-notifications-populate-preferences-on-migrate)

pandafy: Federico was pointing to the heading underline. The length of the underline should be equal to the length of the text. 

```suggestion
``OPENWISP_NOTIFICATIONS_EMAIL_BATCH_INTERVAL``
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

nemesifier (changes requested):
One thing I had asked before, but I am not sure I put in the review, I think we should add a way to open the web UI of openwisp and trigger the opening of the notification widget automatically, so that we can use that in the email: after clicking on the button to see all the remaining notifications, the user is taken to the application and the notification widget opens automatically to show the notifications.

openwisp_notifications/tests/test_notifications.py
```diff
@@ -953,37 +953,50 @@ def test_batch_email_notification_with_descriptions(self, mock_send_email):
...
+        self.notification_options.update(
+            {
+                'message': 'Notification title',
+                'type': 'default',
+            }
+        )
+        for _ in range(3):
+            notify.send(recipient=self.admin, **self.notification_options)
+
+        self.assertEqual(len(mail.outbox), 3)
```
nemesifier, line 998: So, is this working now @Dhanus3133?

openwisp_notifications/tasks.py
```diff
@@ -272,7 +272,7 @@ def send_batched_email_notifications(instance_id):
         extra_context = {}
         if notifications_count > display_limit:
             extra_context = {
-                'call_to_action_url': f"https://{current_site.domain}/admin",
+                'call_to_action_url': f"https://{current_site.domain}/admin?open_notification_widget=true",
```
nemesifier, line 275: Can we use something more user friendly and shorter like `https://{current_site.domain}/admin/#notifications` please?

openwisp_notifications/tests/test_notifications.py
Dhanus3133: No, it's not. I'm not sure why CI for the PR isn't running.

Dhanus3133: Seems Gagan fixed it. It seems it will be triggered at the next commit.

openwisp_notifications/tasks.py
Dhanus3133: Sure, done.

openwisp_notifications/handlers.py
```diff
@@ -25,7 +25,6 @@
 logger = logging.getLogger(__name__)
 
 EXTRA_DATA = app_settings.get_config()['USE_JSONFIELD']
-EMAIL_BATCH_INTERVAL = app_settings.EMAIL_BATCH_INTERVAL
```
nemesifier: This is executed at project startup, during tests it won't be possible to override it (trying to override it would have no effect, because this line was already computed before the override).

openwisp_notifications/handlers.py
```diff
@@ -205,6 +204,7 @@ def send_email_notification(sender, instance, created, **kwargs):
             'email_id': instance.recipient.email,
         },
     )
+    EMAIL_BATCH_INTERVAL = app_settings.EMAIL_BATCH_INTERVAL
```
nemesifier, line 207: This is executed at runtime, therefore we can override `app_settings.EMAIL_BATCH_INTERVAL` during tests with `mock.patch`.

nemesifier (commented):
@Dhanus3133 I sent a fix for the test, see my comments below to understand why these changes were needed.

nemesifier (approved):
LGTM, I think we can merge.

pandafy (commented):
I created 40 notifications and received this email. 

Screenshot from 2024-07-16 14-21-48

Weren't we supposed to show only limited notification in the email?

openwisp_notifications/tasks.py
```diff
@@ -202,3 +211,79 @@ def delete_ignore_object_notification(instance_id):
...
+    current_site = Site.objects.get_current()
+    email_id = cache_data.get('email_id')
+
+    # Send individual email if there is only one notification
+    if notifications_count == 1:
+        notification = unsent_notifications.first()
+        send_notification_email(notification)
+    else:
+        # Show the amount of notifications according to configured display limit
+        show_notification_description = notifications_count <= display_limit
```
nemesifier, line 244: I think we could display the description anyway, but we have to limit the amount of notifications shown

openwisp_notifications/tasks.py
```diff
@@ -202,3 +211,79 @@ def delete_ignore_object_notification(instance_id):
...
+        for notification in unsent_notifications:
```
nemesifier, line 245:
```suggestion
        for notification in unsent_notifications[0:display_limit - 1]
```

nemesifier (changes requested):

openwisp_notifications/templates/emails/batch_email.txt
```diff
@@ -0,0 +1,16 @@
+{% load i18n %}
+
+[{{ site_name }}] {{ notifications_count }} {% translate "new notifications since" %} {{ start_time }}
+
+{% for notification in notifications %}
+- {{ notification.message }}
+  {% translate "URL" %}: {{ notification.url }}
+  {% translate "Timestamp" %}: {{ notification.timestamp|date:"F j, Y, g:i a" }}
+  {% if show_notification_description %}
+      {% translate "Description" %}: {{ notification.rendered_description }}
```
nemesifier, line 10: The description line should be shown after the message and only if not empty, without indentation. Let's always show the description if not empty, regardless of the amount of notifications.

openwisp_notifications/templates/emails/batch_email.txt
```diff
@@ -0,0 +1,16 @@
+{% load i18n %}
+
+[{{ site_name }}] {{ notifications_count }} {% translate "new notifications since" %} {{ start_time }}
+
+{% for notification in notifications %}
+- {{ notification.message }}
+  {% translate "URL" %}: {{ notification.url }}
+  {% translate "Timestamp" %}: {{ notification.timestamp|date:"F j, Y, g:i a" }}
```
nemesifier, line 8: Timestamp should be moved after description and should be named "Date & Time".

openwisp_notifications/templates/emails/batch_email.txt
```diff
@@ -0,0 +1,16 @@
+{% load i18n %}
+
+[{{ site_name }}] {{ notifications_count }} {% translate "new notifications since" %} {{ start_time }}
+
+{% for notification in notifications %}
+- {{ notification.message }}
+  {% translate "URL" %}: {{ notification.url }}
```
nemesifier, line 7: URL should be last.

openwisp_notifications/templates/emails/batch_email.txt
```diff
@@ -0,0 +1,16 @@
...
+  {% translate "Timestamp" %}: {{ notification.timestamp|date:"F j, Y, g:i a" }}
+  {% if show_notification_description %}
+      {% translate "Description" %}: {{ notification.rendered_description }}
+  {% endif %}
+{% endfor %}
+
+{% if not show_notification_description  %}
+{% translate "View all Notifications" %}: {{ call_to_action_url }}
```
nemesifier, line 15: In the test output shown by @pandafy the URL is not showing.

nemesifier:
```suggestion
{{ call_to_action_text }}: {{ call_to_action_url }}
```

nemesifier (changes requested):
We found a few more problems.

pandafy (commented):
```diff
diff --git a/openwisp_notifications/tasks.py b/openwisp_notifications/tasks.py
index cc44d9b..1a71168 100644
--- a/openwisp_notifications/tasks.py
+++ b/openwisp_notifications/tasks.py
@@ -266,15 +266,15 @@ def send_batched_email_notifications(instance_id):
             'site_name': current_site.name,
             'start_time': starting_time,
         }
-        html_content = render_to_string('emails/batch_email.html', context)
-        plain_text_content = render_to_string('emails/batch_email.txt', context)
-
         extra_context = {}
         if notifications_count > display_limit:
             extra_context = {
                 'call_to_action_url': f"https://{current_site.domain}/admin/#notifications",
-                'call_to_action_text': _('View all Notifications'),
+                'call_to_action_text': _('View all notifications'),
             }
+        context.update(extra_context)
+        html_content = render_to_string('emails/batch_email.html', context)
+        plain_text_content = render_to_string('emails/batch_email.txt', context)

         send_email(
             subject=f'[{current_site.name}] {notifications_count} new notifications since {starting_time}',
diff --git a/openwisp_notifications/templates/emails/batch_email.txt b/openwisp_notifications/templates/emails/batch_email.txt
index 20338cb..3eca567 100644
--- a/openwisp_notifications/templates/emails/batch_email.txt
+++ b/openwisp_notifications/templates/emails/batch_email.txt
@@ -12,5 +12,5 @@
 {% endfor %}

 {% if not show_notification_description  %}
-{% translate "View all Notifications" %}: {{ call_to_action_url }}
+{{ call_to_action_text }}: {{ call_to_action_url }}
 {% endif %}

```

openwisp_notifications/tasks.py
Dhanus3133: Updated!

openwisp_notifications/tasks.py
Dhanus3133: Updated

openwisp_notifications/templates/emails/batch_email.txt
Dhanus3133: Updated.

openwisp_notifications/templates/emails/batch_email.txt
Dhanus3133: Updated!

nemesifier (changes requested):
Whitespace looks wrong, please change the template until it looks like the following:

```
- Default notification with default verb and level info by default
  Date & Time: July 17, 2024, 6:10 p.m.
  URL: http://example.com/api/v1/notifications/notification/9dbf11f7-f3ef-43d2-9a72-a678f2150c84/redirect/

- Default notification with default verb and level info by default
  Date & Time: July 17, 2024, 6:10 p.m.
  URL: http://example.com/api/v1/notifications/notification/fbc5874a-d852-4384-94f2-43280841548e/redirect/
```

There should be a blank line only to separate one notification from the other, make sure the indentation is consistent.

openwisp_notifications/templates/emails/batch_email.txt
```diff
@@ -0,0 +1,18 @@
...
+
+{% for notification in notifications %}
+- {{ notification.message }}
+  {% if notification.rendered_description %}
+  {% translate "Description" %}: {{ notification.rendered_description }}
+  {% endif %}
+  {% translate "Date & Time" %}: {{ notification.timestamp|date:"F j, Y, g:i a" }}
+  {% if notification.url %}
+  {% translate "URL" %}: {{ notification.url }}
+  {% endif %}
```
pandafy, line 13:
```suggestion
- {{ notification.message }}{% if notification.rendered_description %}
  {% translate "Description" %}: {{ notification.rendered_description }}{% endif %}
  {% translate "Date & Time" %}: {{ notification.timestamp|date:"F j, Y, g:i a" }}{% if notification.url %}
  {% translate "URL" %}: {{ notification.url }}{% endif %}
```

nemesifier (commented):
I tested the generated HTML and found the following issue:

```html
<p>
<a href="http://example.com/api/v1/notifications/notification/9f244db8-1897-45ed-9f6c-622d45816f82/redirect/" target="_blank">Default notification with default verb and level info by</a>
<a href="http://example.com/admin/openwisp_users/organization/d811aa06-8020-4d32-bdea-fcbd66e96574/change/">default</a>
</p>
```

The second link goes straight to the object without passing through the notification API.

@pandafy wasn't this something that we solved recently?

openwisp_notifications/templates/emails/batch_email.html
```diff
@@ -0,0 +1,107 @@
...
+    <div>
+        {% for notification in notifications %}
+            <div class="alert {{ notification.level }}">
+                <h2>
+                    <span class="badge {{ notification.level }}">{{ notification.level|upper }}</span>
+                    <span class="title">
+                        {% if notification.url %}
+                            <a href="{{ notification.url }}" target="_blank">{{ notification.message }}</a>
+                        {% else %}
+                            {{ notification.message }}
```
pandafy, line 96:
```suggestion
                            {{ notification.email_message }}
```

openwisp_notifications/templates/emails/batch_email.txt
```diff
@@ -0,0 +1,14 @@
+{% load i18n %}
+
+[{{ site_name }}] {{ notifications_count }} {% translate "new notifications since" %} {{ start_time }}
+
+{% for notification in notifications %}
+- {{ notification.message }}{% if notification.rendered_description %}
```
pandafy, line 6:
```suggestion
- {{ notification.email_message }}{% if notification.rendered_description %}
```

openwisp_notifications/utils.py
```diff
@@ -28,3 +32,34 @@ def normalize_unread_count(unread_count):
...
+def send_notification_email(notification):
+    """Send a single email notification"""
+    try:
+        subject = notification.email_subject
+    except NotificationRenderException:
+        # Do not send email if notification is malformed.
+        return
+    url = notification.data.get('url', '') if notification.data else None
+    description = notification.message
```
pandafy, line 45:
```suggestion
    description = notification.email_message
```

openwisp_notifications/utils.py
```diff
@@ -28,3 +32,34 @@ def normalize_unread_count(unread_count):
...
+    else:
+        target_url = None
+    if target_url:
+        description += _('\n\nFor more information see %(target_url)s.') % {
+            'target_url': target_url
+        }
+    send_email(
+        subject,
+        description,
+        notification.message,
```
pandafy, line 59:
```suggestion
        notification.email_message,
```

pandafy (changes requested):
The last assertion in this test looks bogus. It should have caught this issue 

https://github.com/openwisp/openwisp-notifications/blob/3b0fc9df021dcc94cff07a8b51edcd6260473517/openwisp_notifications/tests/test_notifications.py#L313-L345

openwisp_notifications/templates/emails/batch_email.html
Dhanus3133: Updated.

openwisp_notifications/utils.py
Dhanus3133: Updated.

openwisp_notifications/tests/test_notifications.py
```diff
@@ -944,6 +944,61 @@ def test_notification_for_unverified_email(self):
         # we don't send emails to unverified email addresses
         self.assertEqual(len(mail.outbox), 0)
 
+    @patch('openwisp_notifications.tasks.send_batched_email_notifications.apply_async')
+    def test_batch_email_notification(self, mock_send_email):
```
pandafy, line 948: @Dhanus3133 none of these test verifies the actual formatting of the email. We do not have tests for the text email at all.

pandafy (changes requested):

openwisp_notifications/tests/test_notifications.py
Dhanus3133: Updated!

nemesifier (approved):

PR #278, opened 2024-06-04
[fix] Prevent sending multiple identical notifications #277

### LINKED ISSUE #277: [bug] User receives multiple identical notifications
If a user is in several organizations,
they will receive multiple identical notifications.

The reason for this is that the database query, which determines which user should receive the notification, can return a user multiple times.

Steps to replicate the issue:
1. Do a fresh installation of openwisp-controller [as described in the docs](https://github.com/openwisp/openwisp-controller?tab=readme-ov-file#id4) and launch the development server.
2. Re-enable the "default" notification type by commenting out the lines [268](https://github.com/openwisp/openwisp-controller/blob/a4b5d429041ccc2559b77e06afe8404156d765dc/openwisp_controller/config/apps.py#L269) to `272` in `openwisp_controller/config/apps.py`.
3. Log In to the admin site with the user created during installation.
4. Create a second organization.
5. Make the user organization user of the two organizations (`is_admin` is not required).
6. Create a nottification type setting for the type "Default Type" with organization `default`, web notification checked and email notification checked.
8. Then create a notification with the following command: `env/bin/python tests/manage.py create_notification`
9. This user will then receive the notification twice.

kkreitmair:
If a user is in several organizations,
they will receive multiple identical notifications. The reason for this is that the database query,
which determines which user should receive the notification, does a LEFT OUTER JOIN with the "openwisp_users_organizationuser" table. In this table, a user is multiple times listed,
when he is in several organizations.
Therefore the query returns such a user multiple times, which will then receive multiple identical notifications. This commit fixes this bug by adding a distinct to all user queries.

Fixes #277

### DIFF
--- a/openwisp_notifications/handlers.py
+++ b/openwisp_notifications/handlers.py
@@ -114,7 +114,9 @@ def notify_handler(**kwargs):
         # Check if recipient is User, Group or QuerySet
         if isinstance(recipient, Group):
             recipients = recipient.user_set.filter(where_group)
-        elif isinstance(recipient, (QuerySet, list)):
+        elif isinstance(recipient, QuerySet):
+            recipients = recipient.distinct()
+        elif isinstance(recipient, list):
             recipients = recipient
         else:
             recipients = [recipient]
@@ -126,6 +128,7 @@ def notify_handler(**kwargs):
             .order_by('date_joined')
             .filter(where)
             .exclude(not_where)
+            .distinct()
         )
     optional_objs = [
         (kwargs.pop(opt, None), opt) for opt in ('target', 'action_object')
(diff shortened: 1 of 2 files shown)


nemesifier (changes requested):
Thanks @kkreitmair for the patch.

The issue description and patch is missing a few things:

- Can you share a series of exact steps to replicate the issue?
- We'll need to replicate the steps to write a test, ensure it fails without the patch and ensure the test passes with your patch

kkreitmair:
@nemesifier I updated the issue comment by a description how to replicate the issue.
The test will follow soon.

kkreitmair:
@nemesifier I added the test.

In my patch, I naively added also a `.distinct()` to the [QuerySet](https://github.com/openwisp/openwisp-notifications/pull/278/files#diff-5930139e901f3ac3e29b49c3b4e03a54def94c49852ffe74c18b400d5dfa014cR117-R119), which is injected through a parameter of the function.
This can be a convenient functionality, but could be also seen as a bad practice, because normally the caller of the function should care about this.
So the question is if this change should stay or should be removed. What is your opinion on this?

nemesifier (approved):
I am not sure why we didn't spot this before, but thank you very much @kkreitmair for fixing it!

PR #282, opened 2024-06-19
[notifications-docs] Restructured Documentation: New Unified Documentation Website

pandafy:
Related to https://github.com/openwisp/openwisp-docs/issues/107.
These documents are pulled by the logic being worked on in https://github.com/openwisp/openwisp-docs/pull/189.
We are doing similar changes in all modules.


docs/user/sending-notifications.rst
```diff
@@ -0,0 +1,115 @@
...
+        **kwargs
+    )
+
+Since ``openwisp-notifications`` uses ``django-notifications`` under the
+hood, usage of the ``notify signal`` has been kept unaffected to maintain
+consistency with ``django-notifications``. You can learn more about
+accepted parameters from `django-notifications documentation
+<https://github.com/django-notifications/django-notifications#generating-notifications>`_.
+
+Additional ``notify`` Keyword Arguments
```
nemesifier, line 58: Why do we have an additional section for these parameters? Why can't this section be merged with the previous one?

docs/user/sending-notifications.rst
```diff
@@ -0,0 +1,115 @@
...
+``email_subject`` Sets subject of email notification to be sent.
+
+                  Defaults to the notification message.
+``url``           Adds a URL in the email text, eg:
+
+                  ``For more information see <url>.``
+
+                  Defaults to ``None``, meaning the above message would
+                  not be added to the email text.
+``type``          Set values of other parameters based on registered
```
nemesifier, line 72: I think type should be first in this table, it's probably the most important argument, isn't it?.

docs/user/sending-notifications.rst
```diff
@@ -0,0 +1,115 @@
...
+                  `notification types <#notification-types>`_
```
nemesifier, line 73: I think this needs to be updated. Please double check there aren't other occurrences of `<#` around the restructured docs of all the modules.

I did a project wide search in the modules directory and I found quite a few.

docs/user/sending-notifications.rst
```diff
@@ -0,0 +1,115 @@
...
+
+    from openwisp_notifications.signals import notify
+
+    User = get_user_model()
+    Group = load_model("openwisp_users", "Group")
+    admin = User.objects.get(email="admin@admin.com")
+    operators = Group.objects.get(name="Operator")
+
+    notify.send(
+        sender=admin,
```
nemesifier, line 19: Why is this example not mentioning `type`, ``url`` and ``email_subject``?
Those arguments seem the arguments we use most in OpenWISP, so not having them in this main example is puzzling to me.

nemesifier (changes requested):
Coming to `user/sending-notifications.rst` from a link in the Controller docs, I noticed some area for improvements there.

docs/user/sending-notifications.rst
```diff
@@ -78,7 +77,7 @@ Additional ``notify`` Keyword Arguments
 ================= ======================================================
 
 Passing Extra Data to Notifications
------------------------------------
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
nemesifier, line 80: What do you think about adding a call to `notify()` in the following example to make it clearer?

docs/user/sending-notifications.rst
```diff
@@ -18,7 +18,7 @@ Notifications can be created using the ``notify`` signal. Eg:
     notify.send(
         sender=admin,
         recipient=operators,
-        type='default',
+        type="default",
         description="Test Notification",
         verb="Test Notification",
         email_subject="Test Email Subject",
```
nemesifier, line 24: are you sure this example makes perfect sense?

I was thinking we could use the generic message notification as an example which allows sending a custom message, while the example after `register_notification_type` would show how to use the typical case of sending a notification of a type which has already been defined and we don't  need to pass any message or description because it's been already defined in the type. I would like this important concept to be more easier to grasp in this document, I don't think it's clearly explained now.

nemesifier (changes requested):

nemesifier (approved):
Ready!

PR #290, opened 2024-07-15
[feat] Notification Preferences Page

### LINKED ISSUE #110: [feature] Add a dedicated view for managing notification preferences
> Add a dedicated admin view to allow the current logged in user to change their notification settings, we can then add a link like "change notification preferences" somewhere in the notification JS widget, this would be great for UX

From https://github.com/openwisp/openwisp-notifications/pull/69#discussion_r470868296

Depends on https://github.com/openwisp/openwisp-notifications/issues/255.
This new page must not be an admin page, also non staff users must be able to access it.
This page shall use the REST API to fetch and modify notification settings.

Modification of settings must be done only when actual changes are done, so when saving, the code must be smart enough to understand what HTTP requests it must make to change the settings (avoid generating a ton of useless HTTP requests). 

----------------------------------------------

We need to implement a solution that enables users to easily manage their notification preferences. 

**The shortcoming of the current solution**
`NotificationSettings` are shown as an inline to the `UserAdmin` page. When a user is an admin for multiple organizations, or a superuser, the number of notification settings becomes tremendous, thus hurting the performance and usability of the  `UserAdmin` view.  

**Proposed Solution**

- Add a dedicated view for managing users' notification preferences. 
- Add a link to this view in the notification widget. Clicking on this link should open the notification preferences page for the logged-in user.
- Add a link to this view in the `UserAdmin` change page. Clicking on this link should open the notification preferences of the selected user. 
- View Permissions: 
    - Allow superusers to change the notification preferences of any user
    - Allow non superadmin users to change only their notification preferences. 

**Open for Discussion** 
UI:

- In the new view, we can render notification settings in a table (as it is done now). But, we should allow filtering the table, e.g. [OpenWrt's Table of Hardware](https://openwrt.org/toh/start#table_of_hardware). This allows users to quickly find the notification setting. It will be great if we can use autocomplete fields for searching. 
- We can group notification preferences by organizations: 
  E.g.:
```
+---------------------------------------+
| Organization1                         |
+---------------------------------------+
| +---------------------+-----+-------+ |
| | Notification Type 	| Web | Email | |
| +---------------------+-----+-------+ |
| | Ping Problem        | [ ] | [ ]   | |
| | Ping Recovery       | [ ] | [ ]   | |
| +---------------------+-----+-------+ |
+---------------------------------------+
| Organization2                         |
+---------------------------------------+
| +---------------------+-----+-------+ |
| | Notification Type 	| Web | Email | |
| +---------------------+-----+-------+ |
| | Ping Problem        | [ ] | [ ]   | |
| | Ping Recovery       | [ ] | [ ]   | |
| +---------------------+-----+-------+ |
+---------------------------------------+
```

**Implementation details**:

- To account for systems which have a large number of organizations, we shall make the widget capable of loading multiple pages of notification settings

### LINKED ISSUE #148: [feature] Allow to disable notifications for all organizations or keep everything disabled except notifications for specific organizations

### LINKED ISSUE #255: [feature] Add REST API to manage notification preferences of other users
We need a REST API endpoint which will allow administrators to change the notification settings of other users.

The API endpoint path could be: `/api/v1/users/user/{id}/notification-setting/` or if that proves to be hard to implement due to the modular structure of OpenWISP, we could use `/api/v1/notifications/user/{id}/setting/`.

Only the following users can access this endpoint:

- superusers
- org manager (load only notification settings of the orgs managed which match the orgs the user is member of, if the user being viewed is not member of any organization managed by the org manager, return 403)
- authenticated users who match the user id (any user wishing to change their own preferences).

The logic of this API endpoint must be shared with the logic of the existing `/api/v1/notification/user-settings/` and avoid duplication. 

We need to return the organization name in the API output (let's make sure we use `select_related()` to avoid generating too many queries when loading this data from the DB).

Dhanus3133:
The preference page will look like the following screenshot:
image

Closes #110
Closes #148
Closes #255


openwisp_notifications/api/urls.py
```diff
@@ -9,42 +9,46 @@ def get_api_urls(api_views=None):
...
+        # path(
+        #     'user/user-setting/',
+        #     views.notification_setting_list,
+        #     name='notification_setting_list',
+        # ),
+        # path(
+        #     'user/user-setting/<uuid:pk>/',
+        #     views.notification_setting,
+        #     name='notification_setting',
+        # ),
```
nemesifier: I didn't mean to comment out these URLs, we should just flag them as deprecated in the docs.

Dhanus3133: Yeah had already reverted it back.

openwisp_notifications/base/views.py
```diff
@@ -0,0 +1,40 @@
...
+    def get_context_data(self, **kwargs):
+        context = super().get_context_data(**kwargs)
+        user_id = self.kwargs.get('pk')
+
+        if user_id:
+            try:
+                user = User.objects.get(pk=user_id)
+                # Only admin should access other users settings
+                if not self.request.user.is_staff:
+                    raise Http404("You do not have permission to access this page.")
```
nemesifier, line 22: maintain the use of single quotes for consistency please

openwisp_notifications/static/openwisp-notifications/js/settings.js
```diff
@@ -0,0 +1,207 @@
...
+        const data = {};
+        data[checkbox.data('type')] = checkbox.is(':checked');
+        $.ajax({
+            type: 'PATCH',
+            url: '/api/v1/notifications/user/' + userId + '/user-setting/' + checkbox.data('pk') + '/',
+            headers: { 'X-CSRFToken': $('input[name="csrfmiddlewaretoken"]').val() },
+            contentType: "application/json",
+            data: JSON.stringify(data),
+            success: function () {
+                showToast('success', 'Settings updated successfully.');
```
nemesifier, line 115: Please flag as translatable, look for the `gettext` function in this module or other modules.
```suggestion
                showToast('success', gettext('Settings updated successfully.'));
```
Applies to other user facing strings as well.

openwisp_notifications/static/openwisp-notifications/js/settings.js
```diff
@@ -0,0 +1,207 @@
...
+            },
+            error: function () {
+                showToast('error', 'Something went wrong. Please try again.');
```
nemesifier, line 118: Here too.

openwisp_notifications/static/openwisp-notifications/js/settings.js
```diff
@@ -0,0 +1,207 @@
...
+            type: 'POST',
+            url: '/api/v1/notifications/user/' + userId + '/organization/' + organizationId + '/setting/',
+            headers: { 'X-CSRFToken': $('input[name="csrfmiddlewaretoken"]').val() },
+            contentType: "application/json",
+            data: JSON.stringify(data),
+            success: function () {
+                showToast('success', 'Organization settings updated successfully.');
+            },
+            error: function () {
+                showToast('error', 'Something went wrong. Please try again.');
```
nemesifier, line 139: these two too

openwisp_notifications/static/openwisp-notifications/js/settings.js
```diff
@@ -0,0 +1,207 @@
...
+                type: 'POST',
+                url: '/api/v1/notifications/user/' + userId + '/preference/',
+                headers: { 'X-CSRFToken': $('input[name="csrfmiddlewaretoken"]').val() },
+                contentType: "application/json",
+                data: JSON.stringify(data),
+                success: function () {
+                    showToast('success', 'Global settings updated successfully.');
+                },
+                error: function () {
+                    showToast('error', 'Something went wrong. Please try again.');
```
nemesifier, line 175: here too

nemesifier (changes requested):

openwisp_notifications/base/views.py
Dhanus3133: Resolved.

openwisp_notifications/static/openwisp-notifications/js/settings.js
Dhanus3133: Updated.

openwisp_notifications/static/openwisp-notifications/js/settings.js
Dhanus3133: Updated.

openwisp_notifications/static/openwisp-notifications/js/settings.js
Dhanus3133: Updated.

nemesifier (changes requested):
Good progress @Dhanus3133 :+1: 

Please keep in mind that you need to focus on the point of view of the user (user-centered), making sure the UI provides up to date data and guides the user in a way that allows him/her to understand what's going on.

- Web notification on the left, email on the right (as in the current notification settings)
- Keep in mind web notifications are required for email notifications, therefore the web/email checkbox inputs have work as in the current notification settings (if you disable web, email will be also disabled and when both are disabled, if you enable email, web will be also be enabled) 
- Type: show verbose_message in the UI, in the API we call this type_label colors
- Please reuse CSS classes for colors, titles, form row, etc
- Open first org in the UI automatically
- Triangle icon: we should have similar icons to reuse
- Accordion title of the org: reuse the style of other pages, eg:

Screenshot from 2024-08-06 10-18-14

- Table: reuse styles from any change list page (table headings, table body, table rows, checkbox, etc)
- Make sure email and web are <label> tags
- Preferences saved message: please change as discussed (fixed place, different logic than notification toast)
- Look for drop-in replacements of default checkbox inputs to look more similar to what the UX designers shared (but no execute yet, just look for possible open source options to use)

pandafy (changes requested):
Good work @Dhanus3133! In this review, I primarily focused on the UX/UI. I hope these points would help you progress with this PR

- [x] Add link to the notification preference page in the notification widget 
- [x]  Add link to the notification preference page in the UserAdmin
- [x]  When the API request fails, the user operation should rollback. E.g. if the user disabled the global web notifications and the API request fails, then the UI should flag the checkbox as enabled again.
- [x] Add tooltip to the global and organization settings

openwisp_notifications/api/permissions.py
```diff
@@ -0,0 +1,17 @@
...
+
+        # Allow superusers
+        if user.is_superuser:
+            return True
+
+        # Allow users to update their own settings
+        if user.id == user_id:
+            return True
+
+        return False
```
pandafy, line 17: This code can be condensed as 

```suggestion
        return request.user.is_superuser or request.user.id == view.kwargs.get('user_id')
```

Please add a docstring explaining that only superuser can change the notification preferences of other users.

pandafy (changes requested):
In our weekly call, we discussed the following changes to the docs

- explain the new interaction between default notification settings in the type definition and disabled global notifications in the preference of the user
- make sure API docs page is up to date

nemesifier (changes requested):
Notes from today's weekly meeting:

- Handling of API errors: the UI should not wait for succesful response to update itself, it should just rollback in case of errors
- Switch Input Chekbox: we can use this solution (we need to make the :focus more evident and ajust the colors): https://www.w3schools.com/howto/howto_css_switch.asp

nemesifier (changes requested):
- update notification preferences button in user page to link the correct URL
- notification preferences of another user should show the username in the first heading of the page, please do not change the top right header as discussed
- in the notification widget, please change "settings" to "preferences"
- once we are done we will need to update the docs

nemesifier (changes requested):
- Toggle all/Untoggle all button
- Fix conflicts
- Make build pass again
- Update Notification Preferences section in the docs, use a temporary image uploaded anywhere

openwisp_notifications/base/views.py
```diff
@@ -0,0 +1,44 @@
...
+
+    def get_context_data(self, **kwargs):
+        context = super().get_context_data(**kwargs)
+        user_id = self.kwargs.get('pk')
+
+        if user_id:
+            try:
+                user = User.objects.get(pk=user_id)
+                # Only admin should access other users preferences
+                if not self.request.user.is_staff:
```
nemesifier, line 22: Only superadmins are allowed for now.

Hence we should reject `if self.request.user.id != user_id and not user.is_superuser`.

openwisp_notifications/base/views.py
```diff
@@ -0,0 +1,44 @@
...
+                    raise Http404('You do not have permission to access this page.')
```
nemesifier, line 23: 404 is very wrong here. We need to return 403. Please review the HTTP status codes in the RFC.
https://www.rfc-editor.org/rfc/rfc9110.html#name-403-forbidden

openwisp_notifications/templates/openwisp_notifications/preferences.html
```diff
@@ -0,0 +1,59 @@
...
+      <a href="{% url 'admin:index' %}">{% trans 'Home' %}</a> &rsaquo;
+    {% endif %}
+    {% trans 'Notification Preferences' %}
+  </div>
+{% endblock breadcrumbs %}
+
+{% block content %}
+  <div class="settings-container" data-user-id="{{ user.id }}">
+    <div class="global-settings">
+      <strong>Global Settings:</strong>
```
nemesifier, line 32: flag for translation

openwisp_notifications/templates/openwisp_notifications/preferences.html
```diff
@@ -0,0 +1,59 @@
...
+      <div class="global-settings-container">
+        <span>Web</span>
```
nemesifier, line 34: flag for translation

openwisp_notifications/templates/openwisp_notifications/preferences.html
```diff
@@ -0,0 +1,59 @@
...
+        <label class="switch">
+          <input type="checkbox" id="global-web" />
+          <span class="slider round"></span>
+        </label>
+        <span class="tooltip-icon" data-tooltip="{% trans 'Enable or disable web notifications globally' %}">?</span>
+      </div>
+      <div class="global-settings-container">
+        <span>Email</span>
```
nemesifier, line 42: flag for translation

nemesifier (changes requested):
Good progress @Dhanus3133.
The test coverage has decreased significantly.
Please address this and other issues I highlighted below.

nemesifier (changes requested):
I pushed some UI improvements.

Please also rebase on the latest master.

Regaring this image:
Screenshot from 2024-09-02 13-31-35

Please use the gray version for consistency.

openwisp_notifications/static/openwisp-notifications/js/preferences.js
```diff
@@ -0,0 +1,445 @@
...
+(function ($) {
+    let isUpdateInProgress = false;
+
+    $(document).ready(function () {
+        const userId = $('.settings-container').data('user-id');
+        fetchGlobalSettings(userId);
+    });
+
+    function fetchGlobalSettings(userId) {
+        $.getJSON(`/api/v1/notifications/user/${userId}/preference/`, function (globalData) {
```
pandafy, line 16: Can you format the URL as we do it for the notification widget?, i.e prepend the `notificationApiHost`.

https://github.com/openwisp/openwisp-notifications/blob/63a29862e97faa1cc5847a149189be06171ee564/openwisp_notifications/static/openwisp-notifications/js/notifications.js#L316

openwisp_notifications/static/openwisp-notifications/js/preferences.js
```diff
@@ -0,0 +1,445 @@
...
+            showToast('error', gettext('Error fetching global settings. Please try again.'));
+        });
+    }
+
+    function fetchNotificationSettings(userId, isGlobalWebChecked, isGlobalEmailChecked) {
+        let allResults = [];
+        let currentPage = 1;
+
+        (function fetchPage() {
+            $.getJSON(`/api/v1/notifications/user/${userId}/user-setting/?page_size=100&page=${currentPage}`, function (data) {
```
pandafy, line 36: Can you format the URL as we do it for the notification widget?, i.e prepend the `notificationApiHost`.

https://github.com/openwisp/openwisp-notifications/blob/63a29862e97faa1cc5847a149189be06171ee564/openwisp_notifications/static/openwisp-notifications/js/notifications.js#L316

openwisp_notifications/static/openwisp-notifications/js/preferences.js
```diff
@@ -0,0 +1,445 @@
...
+
+            const table = $(this).closest('table');
+            table.find('.web-checkbox').prop('checked', isOrgWebChecked).change();
+            table.find('.email-checkbox').prop('checked', isOrgEmailChecked).change();
+
+            updateMainCheckboxes(table);
+
+            $.ajax({
+                type: 'POST',
+                url: `/api/v1/notifications/user/${userId}/organization/${orgId}/setting/`,
```
pandafy, line 278: Can you format the URL as we do it for the notification widget?, i.e prepend the `notificationApiHost`.

https://github.com/openwisp/openwisp-notifications/blob/63a29862e97faa1cc5847a149189be06171ee564/openwisp_notifications/static/openwisp-notifications/js/notifications.js#L316

pandafy (changes requested):

openwisp_notifications/base/views.py
pandafy, line 1: Can we call this file `openwisp_notifications/views.py`.

openwisp_notifications/urls.py
```diff
@@ -1,6 +1,7 @@
 from django.urls import include, path
 
 from .api.urls import get_api_urls
+from .base.views import notification_preference_page
```
pandafy, line 4: This URL for this view should be added in openwisp_notifications/urls.py and not in the `api` sub-module. 

https://github.com/openwisp/openwisp-notifications/blob/63a29862e97faa1cc5847a149189be06171ee564/openwisp_notifications/urls.py#L6-L15

openwisp_notifications/base/views.py
Dhanus3133: Updated.

openwisp_notifications/static/openwisp-notifications/js/preferences.js
Dhanus3133: Updated.

openwisp_notifications/static/openwisp-notifications/js/preferences.js
Dhanus3133: Updated.

openwisp_notifications/static/openwisp-notifications/js/preferences.js
Dhanus3133: Updated.

openwisp_notifications/templates/openwisp_notifications/preferences.html
Dhanus3133: Updated.

openwisp_notifications/templates/openwisp_notifications/preferences.html
Dhanus3133: Updated.

openwisp_notifications/templates/openwisp_notifications/preferences.html
Dhanus3133: Updated.

openwisp_notifications/base/views.py
Dhanus3133: I am removing this one. Because it's been directly handled in the `test_func`

openwisp_notifications/base/views.py
```diff
@@ -0,0 +1,44 @@
...
+                raise Http404('User does not exist')
+        else:
+            user = self.request.user
+            context['title'] = _('Notification Preference')
+
+        context['user_id'] = user.id
+        return context
+
+    def test_func(self):
+        '''
```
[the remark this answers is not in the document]
Dhanus3133: https://docs.djangoproject.com/en/5.1/topics/auth/default/#django.contrib.auth.mixins.UserPassesTestMixin.test_func

openwisp_notifications/api/urls.py
```diff
@@ -9,32 +9,61 @@ def get_api_urls(api_views=None):
     if not api_views:
         api_views = views
     return [
-        path('', views.notifications_list, name='notifications_list'),
-        path('read/', views.notifications_read_all, name='notifications_read_all'),
-        path('<uuid:pk>/', views.notification_detail, name='notification_detail'),
+        path('notification/', views.notifications_list, name='notifications_list'),
```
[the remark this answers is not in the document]
Dhanus3133: Updated.

openwisp_notifications/api/views.py
```diff
@@ -119,6 +124,14 @@ class BaseNotificationSettingView(GenericAPIView):
     def get_queryset(self):
         if getattr(self, 'swagger_fake_view', False):
             return NotificationSetting.objects.none()  # pragma: no cover
+
+        user_id = self.kwargs.get('user_id')
+
+        if user_id:
+            if not (self.request.user.id == user_id or self.request.user.is_staff):
+                raise PermissionDenied()
```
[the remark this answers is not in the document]
Dhanus3133: Yes. Updated.

openwisp_notifications/api/permissions.py
```diff
@@ -0,0 +1,14 @@
+from rest_framework.permissions import BasePermission
+
+
+class IsAuthenticatedToUpdateNotificationSetting(BasePermission):
+    """
+    Permission class to check if the user is authenticated to update notification settings.
+    Allows superusers to change the notification preferences of other users.
+    Regular users can only update their own settings.
+    """
```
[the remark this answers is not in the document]
Dhanus3133: Updated.

openwisp_notifications/api/permissions.py
```diff
@@ -0,0 +1,14 @@
+from rest_framework.permissions import BasePermission
+
+
+class IsAuthenticatedToUpdateNotificationSetting(BasePermission):
```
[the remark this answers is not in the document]
Dhanus3133: Updated.

openwisp_notifications/templates/openwisp_notifications/preferences.html
```diff
@@ -0,0 +1,57 @@
...
+      <a href="{% url 'admin:app_list' app_label='openwisp_users' %}">{% trans 'Users and Organizations' %}</a> &rsaquo;
+      <a href="{% url 'admin:openwisp_users_user_changelist' %}">{% trans 'Users' %}</a> &rsaquo;
+      <a href="{% url 'admin:openwisp_users_user_change' user_id %}">{{ username }}</a> &rsaquo;
+    {% endif %}
+    {% trans 'Notification Preferences' %}
+  </div>
+{% endblock breadcrumbs %}
+
+{% block content %}
+  <div class="settings-container" data-user-id="{{ user.id }}">
```
pandafy, line 28: This is reading the logged in user's ID.

pandafy:
```suggestion
  <div class="settings-container" data-user-id="{{ user_id }}">
```

Add a failing test for this.

openwisp_notifications/templates/admin/openwisp_users/user/change_form_object_tools.html
```diff
@@ -0,0 +1,8 @@
+{% extends "admin/change_form_object_tools.html" %}
+
+{% load i18n admin_urls %}
+
+{% block object-tools-items %}
+  <li><a class="button" href="{% url 'notifications:user_notification_preference' object_id %}">Notification Preferences</a></li>
```
pandafy, line 6: Let's only show this button on staff users. Notification preferences are not created for non-staff users.

openwisp_notifications/base/views.py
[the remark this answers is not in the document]
nemesifier: This is new for me, thanks.

pandafy (changes requested):
// The feature allowing changes to notification preferences for other users is currently broken. Instead of displaying the target user's preferences, the page incorrectly shows the preferences of the currently logged-in user.

@Dhanus3133, this highlights the crucial need for thorough testing to prevent bugs like this from slipping into a release. While unit tests are valuable for their speed and focus on specific code segments, integration tests are equally important. They ensure that the entire flow is functioning as intended. This bug likely resulted from a minor mistake, perhaps even a typo, but it could have been caught with an integration test that verifies the full flow—from visiting the URL to checking the correct elements in the HTML.

Moving forward, we expect you to prioritize testing more rigorously - both manual and automated.

openwisp_notifications/templates/openwisp_notifications/preferences.html
Dhanus3133: Sorry about that mistake, I'll be more careful next time!

openwisp_notifications/templates/admin/openwisp_users/user/change_form_object_tools.html
Dhanus3133: Updated.

openwisp_notifications/templates/openwisp_notifications/preferences.html
Dhanus3133: Also it's been handled here 

https://github.com/openwisp/openwisp-notifications/blob/7982603fe9e51531b1e12b82318b0eb377bbced8/openwisp_notifications/tests/test_notifications.py#L1074-L1077

openwisp_notifications/templates/admin/openwisp_users/user/change_form_object_tools.html
```diff
@@ -3,6 +3,8 @@
 {% load i18n admin_urls %}
 
 {% block object-tools-items %}
-  <li><a class="button" href="{% url 'notifications:user_notification_preference' object_id %}">Notification Preferences</a></li>
+  {% if request.user.is_staff %}
```
pandafy, line 6: This should not check the logged-in user, but the user object for which the change page is being rendered. The user belonging to the `object_id`.

pandafy (changes requested):

nemesifier (changes requested):
Notes during today's weekly meeting:

- Let's add a selenium test for the notification preferences page which covers the main UI features
- Show notification preferences button in the user page only if:
   - current logged in user is superuser
   - user being viewed is staff
   At a later stage we should improve this aspect as explained in https://github.com/openwisp/openwisp-notifications/issues/312.

openwisp_notifications/static/openwisp-notifications/js/preferences.js
```diff
@@ -0,0 +1,449 @@
...
+(function ($) {
+    let isUpdateInProgress = false;
+
+    $(document).ready(function () {
+        const userId = $('.settings-container').data('user-id');
+        fetchGlobalSettings(userId);
+    });
+
+    function fetchGlobalSettings(userId) {
+        $.getJSON(getAbsoluteUrl(`/api/v1/notifications/user/${userId}/preference/`), function (globalData) {
```
pandafy, line 20: If you use `$.ajax` here, then you use the `beforeSend` callback for showing overlay and `complete` callback for hiding it. 

https://github.com/openwisp/openwisp-controller/blob/8001c3b9fa7783b5e33bfb5057d2cf5f2ce5bf25/openwisp_controller/connection/static/connection/js/commands.js#L348-L355

pandafy (changes requested):
Show a loader while fetching the notification settings from the API. Hide the loader when the API 


I checked the HTML of the notification preferences page and the loading overlay is already present in the page.

openwisp_notifications/api/views.py
```diff
@@ -198,11 +206,64 @@ def perform_create(self, serializer):
...
+            user_id=user_id,
+            organization=None,
+            type=None,
+            defaults={'email': True, 'web': True},
+        )
+        serializer = self.get_serializer(notification_settings)
+        return Response(serializer.data, status=status.HTTP_200_OK)
+
+    def post(self, request, user_id):
+        serializer = self.get_serializer(data=request.data)
```
pandafy, line 240: Can we use the `NotificationSettingSerializer`  also for updating the global notifications?

pandafy (changes requested):

openwisp_notifications/api/views.py
```diff
@@ -114,11 +116,17 @@ class BaseNotificationSettingView(GenericAPIView):
     model = NotificationSetting
     serializer_class = NotificationSettingSerializer
     authentication_classes = [BearerAuthentication, SessionAuthentication]
-    permission_classes = [IsAuthenticated]
+    permission_classes = [PreferencesPermission]
```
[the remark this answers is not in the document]
Dhanus3133: I believe you meant 401, updated.

openwisp_notifications/api/views.py
```diff
@@ -114,11 +116,17 @@ class BaseNotificationSettingView(GenericAPIView):
...
     def get_queryset(self):
         if getattr(self, 'swagger_fake_view', False):
             return NotificationSetting.objects.none()  # pragma: no cover
+
+        user_id = self.kwargs.get('user_id')
+
+        if user_id:
+            return NotificationSetting.objects.filter(user_id=user_id)
+
         return NotificationSetting.objects.filter(user=self.request.user)
```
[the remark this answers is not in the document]
Dhanus3133: Updated.

openwisp_notifications/api/filters.py
```diff
@@ -8,3 +8,8 @@ class NotificationSettingFilter(OrganizationMembershipFilter):
     class Meta(OrganizationMembershipFilter.Meta):
         model = NotificationSetting
         fields = OrganizationMembershipFilter.Meta.fields + ['type']
+
+    @property
+    def qs(self):
+        parent_qs = super().qs
+        return parent_qs.exclude(organization__isnull=True, type__isnull=True)
```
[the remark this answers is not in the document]
Dhanus3133: We already have a seperate endpoint to update the global notifications here

https://github.com/openwisp/openwisp-notifications/blob/9661ae6bcd9af34b4cef4ffc380dd0f0007030dd/openwisp_notifications/api/urls.py#L53-L57

openwisp_notifications/api/views.py
Dhanus3133: You are right, updated.

openwisp_notifications/api/views.py
```diff
@@ -198,11 +206,64 @@ def perform_create(self, serializer):
...
+    serializer_class = NotificationSettingUpdateSerializer
+
+    def post(self, request, user_id, organization_id):
+        notification_settings = NotificationSetting.objects.filter(
+            organization_id=organization_id, user_id=user_id
+        )
+        serializer = self.get_serializer(data=request.data)
+        if serializer.is_valid():
+            for notification_setting in notification_settings:
+                serializer.update(notification_setting, serializer.validated_data)
```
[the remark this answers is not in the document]
Dhanus3133: Thanks, Updated.

openwisp_notifications/api/views.py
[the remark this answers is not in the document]
pandafy: Yes, right. Thank yoU!

nemesifier (changes requested):
Notes from last round of testing during our weekly meeting:

- Fix exception when deleting user (write a failing test first please)
- In the test "test_post_migration_handler", add assertions to verify that global settings are being created
- Add validation in the model to ensure there can be only 1 global setting for each user
- Let's use only 1 endpoint for notification settings
- Remove the CSS transition from the arrow in the accordion of the UI

pandafy (commented):
Good progress @Dhanus3133 👏🏼 With these last few improvements, we will get this PR ready to merge.


@nemesifier what will be the best way to convey that there was an error loading the notification settings. There's a toast message which notifies that there was an error but the toast is not persistent. Shall we add a paragraph element conveying that there was an error?

Screenshot from 2024-09-16 23-03-27

openwisp_notifications/tests/test_admin.py
```diff
@@ -159,56 +157,6 @@ def test_websocket_protocol(self):
...
-
-        with self.subTest('Test for non-superuser'):
-            self.assertFalse(
-                self.ns_inline.has_change_permission(op_request),
-            )
-            self.assertTrue(
-                self.ns_inline.has_change_permission(op_request, obj=op_request.user),
-            )
-
     def test_org_admin_view_same_org_user_notification_setting(self):
```
pandafy, line 160: You can also remove this test.

openwisp_notifications/tests/test_api.py
```diff
@@ -544,19 +546,20 @@ def test_notification_setting_list_api(self):
...
             self.assertIn('id', notification_setting)
-            self.assertIsNone(notification_setting['web'])
-            self.assertIsNone(notification_setting['email'])
+            self.assertTrue(notification_setting['web'])
+            self.assertTrue(notification_setting['email'])
             self.assertIn('organization', notification_setting)
 
     def test_list_notification_setting_filtering(self):
         url = self._get_path('notification_setting_list')
+        tester = self._create_user()
```
pandafy, line 562: Can you instead create an organization admin here? It will make the test more realistic.

openwisp_notifications/tests/test_api.py
```diff
@@ -592,8 +595,40 @@ def test_list_notification_setting_filtering(self):
...
+            self.client.force_login(self.admin)
+            user_url = self._get_path('user_notification_setting_list', tester.pk)
+            response = self.client.get(user_url)
+            self.assertEqual(response.status_code, 200)
+
+        with self.subTest('Test with user_id by user_id as the same user'):
+            self.client.force_login(tester)
+            user_url = self._get_path('user_notification_setting_list', tester.pk)
+            response = self.client.get(user_url)
+            self.assertEqual(response.status_code, 200)
```
pandafy, line 614: And then verify that the number of notification setting is correct here. 

```
NotificationSetting.objects.filter(user=tester).count()
```

openwisp_notifications/tests/test_api.py
```diff
@@ -592,8 +595,40 @@ def test_list_notification_setting_filtering(self):
...
-        notification_setting = NotificationSetting.objects.first()
+        notification_setting = NotificationSetting.objects.exclude(
+            organization__isnull=True
+        ).first()
+        tester = self._create_user()
+        tester_notification_setting = NotificationSetting.objects.create(
+            user=tester,
+            type='default',
+            organization=Organization.objects.first(),
+        )
```
pandafy, line 631: You wouldn't need to create the notification setting manually if you create an organization admin here instead.

openwisp_notifications/tests/test_api.py
```diff
@@ -671,6 +798,54 @@ def _unread_notification(notification):
...
+            url = self._get_path(
+                'organization_notification_setting', self.admin.pk, org.pk
+            )
+            response = self.client.post(url, data={'web': True, 'email': True})
+            self.assertEqual(response.status_code, 200)
+            notification_setting = NotificationSetting.objects.filter(
+                user=self.admin, organization_id=org.pk
+            ).first()
+            self.assertTrue(notification_setting.web, True)
+            self.assertTrue(notification_setting.email, True)
```
pandafy, line 815: We should assert that all notification settings of the user for this organization was changed. 

```suggestion
        with self.subTest('Test for current user'):
            url = self._get_path(
                'organization_notification_setting', self.admin.pk, org.pk
            )
            NotificationSetting.objects.filter(user=self.admin, organization_id=org.pk).update(email=False, web=False)
            org__setting_count = NotificationSetting.objects.filter(user=self.admin, organization_id=org.pk).count()
            response = self.client.post(url, data={'web': True, 'email': True})
            self.assertEqual(response.status_code, 200)
            self.assertEqual(NotificationSetting.objects.filter(user=self.admin, organization_id=org.pk, email=True, web=True).count(), org_setting_count)
```

openwisp_notifications/tests/test_api.py
```diff
@@ -671,6 +798,54 @@ def _unread_notification(notification):
...
```
pandafy, line 827: Isn't this and the previous sub-test are same?

openwisp_notifications/tests/test_selenium.py
pandafy, line 1: Can you also move the tests present in test_widget.py to this file?

openwisp_notifications/tests/test_selenium.py
```diff
@@ -0,0 +1,69 @@
...
+        self.admin = self._create_admin(
+            username=self.admin_username, password=self.admin_password
+        )
+        self.driver = self.web_driver
+        org = self._create_org()
+        OrganizationUser.objects.create(user=self.admin, organization=org)
+
+    def open(self, url, driver=None):
+        driver = self.driver or self.web_driver
+        driver.get(f'{self.live_server_url}{url}')
```
pandafy, line 30: This is already defined in `SeleniumTestMixin`

openwisp_notifications/tests/test_selenium.py
```diff
@@ -0,0 +1,69 @@
...
+    TestOrganizationMixin,
+    StaticLiveServerTestCase,
+):
+    serve_static = True
+
+    def setUp(self):
+        self.admin = self._create_admin(
+            username=self.admin_username, password=self.admin_password
+        )
+        self.driver = self.web_driver
```
pandafy, line 24: Why?

openwisp_notifications/tests/test_selenium.py
```diff
@@ -0,0 +1,69 @@
...
+        OrganizationUser.objects.create(user=self.admin, organization=org)
+
+    def open(self, url, driver=None):
+        driver = self.driver or self.web_driver
+        driver.get(f'{self.live_server_url}{url}')
+
+    def test_notification_preference_page(self):
+        self.login()
+        self.open('/notifications/preferences/')
+        self.driver.implicitly_wait(1)
```
pandafy, line 35: Why is this required? Isn't `presence_of_element_located` sufficient?

openwisp_notifications/tests/test_utils.py
```diff
@@ -105,12 +105,3 @@ def run_check():
...
                 self.assertIn(error_message, error.hint)
-
-        with self.subTest('Test setting dotted path is not subclass of ModelAdmin'):
-            path = 'openwisp_notifications.admin.NotificationSettingInline'
-            with patch.object(app_settings, 'IGNORE_ENABLED_ADMIN', [path]):
-                error_message = (
-                    f'"{path}" does not subclasses "django.contrib.admin.ModelAdmin"'
-                )
-                error = run_check()
-                self.assertIn(error_message, error.hint)
```
pandafy, line 116: Instead of removing this test, change the path to some other inline class in OpenWISP

`openwisp_users.admin.OrganizationUserInline`

openwisp_notifications/views.py
```diff
@@ -0,0 +1,45 @@
...
+from django.http import Http404
+from django.utils.translation import gettext as _
+from django.views.generic import TemplateView
+
+User = get_user_model()
+
+
+class NotificationPreferencePage(LoginRequiredMixin, UserPassesTestMixin, TemplateView):
+    template_name = 'openwisp_notifications/preferences.html'
+    login_url = '/admin/login/'
```
pandafy, line 12:
```suggestion
    login_url = reverse_lazy('admin:login')
```

openwisp_notifications/tests/test_selenium.py
```diff
@@ -0,0 +1,69 @@
...
+            self.assertTrue(checkbox.is_selected())
+
+        # Check a single email checkbox
+        first_org_email_checkbox = WebDriverWait(self.driver, 10).until(
+            EC.presence_of_element_located((By.ID, 'org-1-email-1'))
+        )
+        first_org_email_checkbox.click()
+        self.assertTrue(
+            first_org_email_checkbox.find_element(By.TAG_NAME, 'input').is_selected()
+        )
```
pandafy, line 69: Also test the scenario where the application returns empty list for the user.

openwisp_notifications/templates/admin/openwisp_users/user/change_form_object_tools.html
```diff
@@ -0,0 +1,12 @@
+{% extends "admin/change_form_object_tools.html" %}
+
+{% load i18n admin_urls %}
+
+{% block object-tools-items %}
+  {% if request.user.is_staff and original.is_staff %}
```
pandafy, line 6: We need a test to ensure that this button only appears on User page of staff users. You don't need to a selenium test, you can achieve this with Django test. Use the `self.assertContains` to check the anchor element is present.

pandafy (changes requested):

openwisp_notifications/base/models.py
```diff
@@ -288,9 +288,36 @@ def __str__(self):
...
+                .exclude(pk=self.pk)
+                .exists()
+            ):
+                raise ValidationError("There can only be one global setting per user.")
+
     def save(self, *args, **kwargs):
+        self.validate_global_setting()
         if not self.web_notification:
             self.email = self.web_notification
+        if not self.organization and not self.type:
```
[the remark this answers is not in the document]
Dhanus3133: So, when we update the global setting, let's say in the Notification preferences UI or unsubscribe for email notifications, this ensures that all the notification settings for the user are toggled even.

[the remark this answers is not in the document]
pandafy: I commented this code and none of the the test failed. Let's discuss about this in our weekly meeting today.

openwisp_notifications/tests/test_api.py
Dhanus3133: Should probably remove it.

openwisp_notifications/api/views.py
```diff
@@ -198,11 +201,40 @@ def perform_create(self, serializer):
...
+                    notification_setting.web = web
+                else:
+                    notification_setting.web = web
+                    notification_setting.email = email
+
+            NotificationSetting.objects.bulk_update(
+                notification_settings, ['web', 'email']
+            )
+            return Response(status=status.HTTP_200_OK)
+        return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
```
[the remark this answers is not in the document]
pandafy: After another review, I came to conclusion that we don't need to make any changes to the serializer. Please make changes to the test for this endpoint as requested in my previous review. We will see if something breaks because this change.

openwisp_notifications/base/models.py
```diff
@@ -277,14 +280,45 @@ class Meta:
...
+                    self.user.notificationsetting_set.exclude(pk=self.pk).update(
+                        web=self.web
+                    )
+                else:
+                    self.user.notificationsetting_set.exclude(pk=self.pk).update(
+                        web=self.web, email=self.email
+                    )
+            except self.__class__.DoesNotExist:
+                pass
         return super().save(*args, **kwargs)
```
pandafy, line 322: What are the drawbacks for upgrading the NotificationSetting using `update()` 

```python
self.user.notificationsetting_set(web=self.web, email=self.email)
```

What case would be left uncovered? Can you demonstrate that with a failing test case?

pandafy, line 322:
```diff
diff --git a/openwisp_notifications/base/models.py b/openwisp_notifications/base/models.py
index 1b441a3..b4bd571 100644
--- a/openwisp_notifications/base/models.py
+++ b/openwisp_notifications/base/models.py
@@ -1,6 +1,6 @@
 import logging
 from contextlib import contextmanager
-
+from django.db import transaction
 from django.conf import settings
 from django.contrib.contenttypes.fields import GenericForeignKey
 from django.contrib.contenttypes.models import ContentType
@@ -303,23 +303,15 @@ class AbstractNotificationSetting(UUIDModel):
                 raise ValidationError("There can only be one global setting per user.")
 
     def save(self, *args, **kwargs):
-        self.validate_global_setting()
+        # Turn off email notification if web notification is off
         if not self.web_notification:
             self.email = self.web_notification
-        if not self.organization and not self.type:
-            try:
-                original = self.__class__.objects.get(pk=self.pk)
-                if self.web and (self.email == original.email):
-                    self.user.notificationsetting_set.exclude(pk=self.pk).update(
-                        web=self.web
-                    )
-                else:
-                    self.user.notificationsetting_set.exclude(pk=self.pk).update(
-                        web=self.web, email=self.email
-                    )
-            except self.__class__.DoesNotExist:
-                pass
-        return super().save(*args, **kwargs)
+        with transaction.atomic():
+            if not self.organization and not self.type:
+                # This is the global setting.
+                # Update all user notification settings
+                self.user.notificationsetting_set.exclude(pk=self.pk).update(web=self.web, email=self.email)
+            return super().save(*args, **kwargs)
 
     def full_clean(self, *args, **kwargs):
         if self.email == self.type_config['email_notification']:

```

pandafy (changes requested):

openwisp_notifications/tests/test_notifications.py
```diff
@@ -944,6 +950,24 @@ def test_notification_for_unverified_email(self):
         # we don't send emails to unverified email addresses
         self.assertEqual(len(mail.outbox), 0)
 
+    def test_validate_global_notification_setting(self):
```
pandafy, line 953: Move this test to test_notification_settings.py

openwisp_notifications/base/models.py
Dhanus3133: I have added a failing testcase for this 

https://github.com/openwisp/openwisp-notifications/blob/a796372fac040afdda7b314f4d491f6eb6167e71/openwisp_notifications/tests/test_api.py#L850-L870

Also did add for the global setting here

https://github.com/openwisp/openwisp-notifications/blob/a796372fac040afdda7b314f4d491f6eb6167e71/openwisp_notifications/tests/test_notification_setting.py#L292-L307

Let me know what do you think about this.

openwisp_notifications/static/openwisp-notifications/js/preferences.js
```diff
@@ -0,0 +1,482 @@
...
+            const isGlobalEmailChecked = globalSetting.email;
+            globalSettingId = globalSetting.id;
+
+            $('#global-web').prop('checked', isGlobalWebChecked);
+            $('#global-email').prop('checked', isGlobalEmailChecked);
+
+            initializeGlobalSettingsEventListener(userId);
+        }
+
+        const groupedData = groupBy(filteredResults, 'organization_name');
```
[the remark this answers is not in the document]
Dhanus3133: I have updated it to group by the `id` (`organization`)

openwisp_notifications/tests/test_api.py
```diff
@@ -671,6 +804,71 @@ def _unread_notification(notification):
...
+            ).update(email=True)
+
+            response = self.client.post(url, data={'web': True, 'email': False})
+
+            self.assertTrue(
+                NotificationSetting.objects.filter(
+                    user=self.admin, organization_id=org.pk, type='default', email=True
+                ).exists()
+            )
+
```
pandafy, line 871: This is the scenario which I think this test emulates: 

Starting state:
```python
            NotificationSetting.objects.filter(
                user=self.admin, organization_id=org.pk, type='default'
            ).update(email=True)

```

Screenshot from 2024-09-18 20-12-10
state: 

Turning off email notifications for the organization: 
```python 
response = self.client.post(url, data={'web': True, 'email': False})
```

Screenshot from 2024-09-18 20-14-40

**Contradictory to the UI**

```python
            self.assertTrue(
                NotificationSetting.objects.filter(
                    user=self.admin, organization_id=org.pk, type='default', email=True
                ).exists()
            )
```

The UI shows that all email notifications for the organization is disabled, which is correct.

openwisp_notifications/base/models.py
pandafy: I commented out the code in question, but the added test didn't fail. :(

I was expecting the test to fail without the added logic.

pandafy (commented):
@Dhanus3133 update the code (including tests) to call the `full_clean` model method before calling `save` on an individual object. 

Django does not perform any validation if you save the object directly. The validation is performed in `full_clean`, hence it is required to call the method before `save`.

docs/user/notification-preferences.rst
```diff
@@ -1,8 +1,8 @@
 Notification Preferences
 ========================
 
-.. image:: https://raw.githubusercontent.com/openwisp/openwisp-notifications/docs/docs/images/notification-settings.png
-    :target: https://raw.githubusercontent.com/openwisp/openwisp-notifications/docs/docs/images/notification-settings.png
+.. image:: https://i.imgur.com/lIGqry5.png
+    :target: https://i.imgur.com/lIGqry5.png
```
nemesifier, line 5: If you guys are fine with the final tweaks, it's time to update this!

docs/user/notification-preferences.rst
```diff
@@ -12,6 +12,10 @@ organized by notification type and organization, allowing users to tailor
 their notification experience by opting to receive updates only from
 specific organizations or notification types.
 
+Users can access and manage their notification preferences by visiting the
+``/notification/preferences/``. Alternatively, this page can also be
+accessed directly from the notification widget.
+
```
nemesifier, line 18: Please add a screenshot here with a red circle on the button.

Let's also invert the order of these sentences, mention the notification preferences button first, which is the most logical way for the user to open this page (nobody will type the URL in their browser, come on).

nemesifier (changes requested):
@pandafy @Dhanus3133 I did some final tweaks!

I noticed an issue in my local env, if I run `makemigrations` locally:

```
./manage.py makemigrations
```

A new file `migrations/0009_alter_notification_options_and_more.py` is created.

Can you check if you get the same behavior?

The CI build isn't giving any warning/error for this but it's better to triple check!

After this we should finally be able to merge!

Dhanus3133:
> A new file `migrations/0009_alter_notification_options_and_more.py` is created.
> 
> Can you check if you get the same behavior?

I tried on my end, but no new migration file was created locally.

openwisp_notifications/views.py
```diff
@@ -0,0 +1,46 @@
...
+        context['title'] = _('Notification Preferences')
+
+        if user_id:
+            try:
+                user = User.objects.only('id', 'username').get(pk=user_id)
+                # Only admin should access other users preferences
+                context['username'] = user.username
+                context['title'] += f' ({user.username})'
+            except User.DoesNotExist:
+                raise Http404('User does not exist')
```
nemesifier, line 27: Flag as translatable

nemesifier (changes requested):

docs/user/notification-preferences.rst
Dhanus3133: Updated

openwisp_notifications/views.py
Dhanus3133: Done

openwisp_notifications/api/urls.py
```diff
@@ -9,32 +9,56 @@ def get_api_urls(api_views=None):
...
+            name='organization_notification_setting',
+        ),
+        # DEPRECATED
+        path(
+            'user/user-setting/',
+            views.notification_setting_list,
+            name='notification_setting_list',
+        ),
+        path(
+            'user/user-setting/<uuid:pk>/',
```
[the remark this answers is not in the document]
Dhanus3133: Yes, I think we have missed it. I have updated now as per current state of the `api/urls.py`.

docs/user/notification-preferences.rst
```diff
@@ -12,6 +12,10 @@ organized by notification type and organization, allowing users to tailor
 their notification experience by opting to receive updates only from
 specific organizations or notification types.
 
+Users can access and manage their notification preferences by visiting the
```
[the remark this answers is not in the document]
nemesifier: > What do you think about adding like this?
> 
> ```
> .. note::
> 
>     - You can disable notifications globally while still enabling them for specific organizations.
>     - Notification settings are now linked: disabling web notifications will automatically disable email notifications, and enabling email notifications will automatically enable web notifications.
>     - Deleting notification settings is no longer possible via the web interface (please use the REST API if removal is needed).
> ```

@Dhanus3133 it's fine, please make sure your local openwisp-utils is up to date and run openwisp-qa-format to reformat the docs changes (we now have a tool like black for restructured text files).

docs/user/notification-preferences.rst
nemesifier: @Dhanus3133 :point_up: this needs to be update as well.

nemesifier (commented):
@Dhanus3133 the image looks fine, but we need to update the main screenshot which is outdated. We also need to push these new screenshots to the docs branch to stop using imgur which was a temporary solution.

PS: let's avoid overwriting existing images in the docs branch, as it may cause images for docs aimed for previous versions to change as well. The best is to create an image in a directory called `25/`.

docs/user/notification-preferences.rst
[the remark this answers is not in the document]
Dhanus3133: Updated.

nemesifier (approved):

PR #333, opened 2025-02-05
[feature] Added warning and send verification email status #281

### LINKED ISSUE #281: [feature] Add warning for email verification status and allow to resend verification email
If a user is accessing the admin and has email notifications enabled, we should check whether the user has its email verified.
If the email is not verified, we can show a warning message using the django message framework, something like:

Email notifications are enabled for your account, but since your email address has not been verified, email sending is currently disabled. Please verify your email address to enable email notifications.

We should link to a view that triggers the resending of the verification email.

SitaGanesh:
## Checklist

Closes #281 

## Description of Changes
**apps.py**

Purpose: Ensure that the signal handlers are registered when the application starts.
Change: Imported `openwisp_notifications.signals.`
Reason: This import ensures that the signal handlers defined in `signals.py` are registered when the application starts. This is necessary for the `user_logged_in` and `post_save signals` to work correctly.

**signals.py**

Purpose: Define signal handlers to manage user profile creation and email verification status.
Changes:
`create_or_save_user_profile:` Ensures that a UserProfile is created or updated whenever a user is created or saved.
`check_user_email_verified:` Checks if the user's email is verified when they log in and displays a warning message if it is not.
Reason: These signal handlers help manage the creation of `user-profiles` and ensure users are informed about their email verification status.

**models.py**

Purpose: Define the `UserProfile` model to track email verification status.
Change: Added the` UserProfile` model with a needs_verification field.
Reason: The UserProfile model is used to store additional information about the user, specifically whether their email needs verification.

**views.py**

Purpose: Provide views to handle email verification and resending verification emails.
Changes:
`resend_verification_email`: Allows users to resend the verification email.
`verify_email`: Handles the verification of the user's email.
Reason: These views enable users to verify their email addresses and resend verification emails if needed.

**urls.py** 

Purpose: Define URL patterns for the new views.
Changes: Added URL patterns for `resend_verification_email` and `verify_email.`
Reason: These URL patterns allow the application to handle requests for resending verification emails and verifying emails.

**urls.py (openwisp2)**

Purpose: Include the new URL patterns in the main URL configuration.
Changes: Included the new URL patterns for `openwisp_notifications`
Reason: This ensures that the new views are accessible through the main application URLs.


openwisp_notifications/models.py
```diff
@@ -26,3 +28,8 @@ class Meta(AbstractIgnoreObjectNotification.Meta):
         swappable = swappable_setting(
             'openwisp_notifications', 'IgnoreObjectNotification'
         )
+
+
+class UserProfile(models.Model):
+    user = models.OneToOneField(settings.AUTH_USER_MODEL, on_delete=models.CASCADE)
+    needs_verification = models.BooleanField(default=False)
```
nemesifier, line 35: Why are you adding this model? There's already a model for EmailAddress from django-allauth, check the openwisp-users module. Are you 100% sure this is necessary? I don't remember asking this in the issue description.

nemesifier (changes requested):
In the issue description I was advising to implement this quite differently than how it is done here.

I was advising to use the [user_logged_in signal](https://docs.djangoproject.com/en/4.2/ref/contrib/auth/#django.contrib.auth.signals.user_logged_in), to which we can add a message using the [message framework](https://docs.djangoproject.com/en/4.2/ref/contrib/messages/).

I'd do something like:

- user logs in, signal receiver is triggered
- make sure the receiver function is executed only if we are dealing with the admin site and/or a staff user (we have to verify whether we can determine if it's the admin site)
- check if the user has any verified email address, eg: `user.emailaddress_set.filter(verified=True).exists()`, this tells you if the user has any verified email address.
- if the user doesn't have any verified email address, show a warning using the message framework, let's make sure we provide a link to resend the verification link via email in the message so they can just click on it to send it

SitaGanesh:
I apologize @nemesifier sir for making these changes incorrectly. I initially misunderstood the implementation approach. Now that I have worked through it, I realize I need further clarification on some points.

**Message Placement:**

- As per the issue description, the warning message should be displayed when a staff user logs in and has email notifications enabled but no verified email address.

- I plan to place the message in base_site.html along with a "Verify Now" link. Could you confirm if this is the correct placement, or should it be restricted to admin templates only?
**Email Verification Flow:**

- Since we are using allauth's send_email_confirmation, the verification link redirects users to allauth’s default confirmation page.
- Would it be acceptable to use this as-is, even if it doesn't match the admin's styling, or should we create a custom view to redirect users back to the admin panel?

**Handling Existing Email Records:**

- I checked the get_or_create method in the resend view, and I want to confirm whether it properly handles cases where the user's email already exists in EmailAddress but remains unverified. Should we account for any edge cases here?
**Django Email Backend:**

- I came across the following setting in settings.py:

`EMAIL_BACKEND = 'django.core.mail.backends.console.EmailBackend'`

- This prints emails to the terminal. To actually send emails, we need to switch to SMTP (Django's default). Should I modify this, or is there any other approach preferred for handling email verification in OpenWISP?



Additionally, I noticed that the issue description suggests using the user_logged_in signal to trigger this message. The steps mentioned include:

- Ensuring the signal runs only for staff users or within the admin panel.
- Checking if the user has a verified email using
`user.emailaddress_set.filter(verified=True).exists()
`
If no verified email exists, display a warning with a link to resend the verification email.


I also saw a mention of adding a UserProfile model with a needs_verification field. Since django-allauth already provides an EmailAddress model, I want to confirm if this addition is necessary. From my understanding, OpenWISP's openwisp-users module already integrates email verification. Please let me know if I missed something.

Would appreciate your guidance on these points!

nemesifier:
> I apologize @nemesifier sir for making these changes incorrectly. I initially misunderstood the implementation approach. Now that I have worked through it, I realize I need further clarification on some points.
> 
> **Message Placement:**
> 
> * As per the issue description, the warning message should be displayed when a staff user logs in and has email notifications enabled but no verified email address.
> * I plan to place the message in base_site.html along with a "Verify Now" link. Could you confirm if this is the correct placement, or should it be restricted to admin templates only?

If you use the [Django Message Framework](https://docs.djangoproject.com/en/4.2/ref/contrib/messages/), you shouldn't need to edit the base template files, which BTW, are placed in openwisp-utils and not in this module.

Have a look around the codebase:

- [warnings](https://github.com/search?q=org%3Aopenwisp+messages.warning%28&type=code)
- [errors]: https://github.com/search?q=org%3Aopenwisp+messages.error%28&type=code

> **Email Verification Flow:**
> 
> * Since we are using allauth's send_email_confirmation, the verification link redirects users to allauth’s default confirmation page.
> * Would it be acceptable to use this as-is, even if it doesn't match the admin's styling, or should we create a custom view to redirect users back to the admin panel?

AFAIK we have already restyled the email confirmation page in openwisp-utils or openwisp-users (can't remember from the top of my head), there shouldn't be need to do anything here.

> **Handling Existing Email Records:**
> 
> * I checked the get_or_create method in the resend view, and I want to confirm whether it properly handles cases where the user's email already exists in EmailAddress but remains unverified. Should we account for any edge cases here?

Please do your own research on this. We can't give you precise instructions on everything otherwise it would take less time to do it ourselves than helping you do it. Test it and find out if there's any edge case which could be problematic for users, if there's any, try to fix it and let us know about it.

>   **Django Email Backend:**
> * I came across the following setting in settings.py:
> 
> `EMAIL_BACKEND = 'django.core.mail.backends.console.EmailBackend'`
> 
> This prints emails to the terminal. To actually send emails, we need to switch to SMTP (Django's default). Should I modify this, or is there any other approach preferred for handling email verification in OpenWISP?

You don't need to send emails during development, the console backend is appropiate here.

> Additionally, I noticed that the issue description suggests using the user_logged_in signal to trigger this message. The steps mentioned include:
> 
> * Ensuring the signal runs only for staff users or within the admin panel.
> * Checking if the user has a verified email using
>   `user.emailaddress_set.filter(verified=True).exists() `
>   If no verified email exists, display a warning with a link to resend the verification email.

Correct.

> I also saw a mention of adding a UserProfile model with a needs_verification field. Since django-allauth already provides an EmailAddress model, I want to confirm if this addition is necessary. From my understanding, OpenWISP's openwisp-users module already integrates email verification. Please let me know if I missed something.

Correct, as already stated, the addition of UserProfile model is not necessary and I don't understand where did you read that requirement? There's no mention of it in the original issue: https://github.com/openwisp/openwisp-notifications/issues/281.

nemesifier:
Here's a few corner cases I just thought about just now.

The user can have multiple email addresses stored.
For resending the verification link, select the primary email address, if available, otherwise the last email address added.
If the user for some reason doesn't have any email address stored in the `EmailAddress` model, but has an email address in the User model, create it and set it as primary.
If the user for some reason doesn't have any email address at all (nor in EmailAddress nor in User model), don't do anything, do not even show the message in the first place.

SitaGanesh:
Sir @nemesifier Thank you for the detailed feedback. I appreciate your patience and guidance.

**UserProfile Model Removal**
I'll remove the UserProfile model entirely. I now understand django-allauth's EmailAddress model already handles verification status. This was my misunderstanding of the original issue requirements.

**Message Placement Correction**
I'll implement the warning using Django's message framework without modifying base_site.html. The message will appear automatically in the admin UI since staff users log in to the admin. I'll verify this works with the default admin templates.

**Email Verification Flow**
I'll retain allauth's confirmation page since OpenWISP already handles styling. No custom views are needed.

**Corner Cases Implementation**
I will need some more time to figure out what actually goes in the code.


> Correct, as already stated, the addition of UserProfile model is not necessary and I don't understand where did you read that requirement? There's no mention of it in the original issue: https://github.com/openwisp/openwisp-notifications/issues/281.

Understood—I initially created a custom UserProfile model with a needs_verification field because I thought it was necessary to track email verification status. However, after your clarification and upon further review of the environment, I observed that django‑allauth’s EmailAddress model (integrated via openwisp‑users) already provides the required functionality. I realize now that I could have extended or utilized the existing models rather than creating my own. I've removed the custom UserProfile model from my implementation accordingly.

nemesifier (changes requested):
Closing this in favour of https://github.com/openwisp/openwisp-notifications/pull/336, next time please push to the same branch to continue the discussion in the same thread.

SitaGanesh:
> Closing this in favour of #336, next time please push to the same branch to continue the discussion in the same thread.

Ok, I got it, sir

PR #344, opened 2025-03-24
[feature] Improve UX of the Notifications Module

### LINKED ISSUE #110: [feature] Add a dedicated view for managing notification preferences

### LINKED ISSUE #132: [feature] Batch email notifications to prevent email flooding

### LINKED ISSUE #148: [feature] Allow to disable notifications for all organizations or keep everything disabled except notifications for specific organizations

### LINKED ISSUE #255: [feature] Add REST API to manage notification preferences of other users

### LINKED ISSUE #306: [docs] Notification batching restructured docs
As discussed:

- Move the relevant settings to the new settings page in `docs/user/settings.rst`
- Add a new page which describes the behavior of the new feature in `docs/user/`
- Add a line in the features intro page that links to the new page 

The docs can be tested via the [docs repo using SSH=1](https://github.com/openwisp/openwisp-docs?tab=readme-ov-file#building-with-ssh-remotes).

pandafy:
Closes #132 
Closes #306
Closes #110 
Closes #148 
Closes #255


.prettierrc.js
pandafy, line 1: This file does not exist in any other OpenWISP module, hence I removed it from here for consistency. 

@Dhanus3133 let us know if this is required.

docs/user/web-email-notifications.rst
```diff
@@ -47,10 +47,58 @@ notification toast.
...
 Along with web notifications OpenWISP Notifications also sends email
 notifications leveraging the :ref:`send_email feature of OpenWISP Utils
 <utils_send_email>`.
+
+.. _notifications_batches:
+
+Email Batches
+~~~~~~~~~~~~~
+
+.. figure:: https://i.imgur.com/W5P009W.png
```
Dhanus3133, line 63: We should update this!

pandafy: Thanks for catching this! I have updated the docs.

pandafy:
@Dhanus3133 can you ensure that the coverage does not decrease for these files: 

- 	api/serializers.py
- base/models.py
- tasks.py

Dhanus3133:
@pandafy For the coverage missing on base/models.py https://github.com/openwisp/openwisp-notifications/blob/b7dff784fd7d1a760202acc4b32c7d04083018f9/openwisp_notifications/base/models.py#L302-L304

Seems to be due to the defaults being set here

https://github.com/openwisp/openwisp-notifications/blob/941982dcb280f0945578da143871e5636354f351/openwisp_notifications/tasks.py#L88-L107

But if that's the case https://github.com/openwisp/openwisp-notifications/blob/b7dff784fd7d1a760202acc4b32c7d04083018f9/openwisp_notifications/base/models.py#L308-L310 this should also cause a coverage issue but it doesn't, not sure why.

nemesifier (changes requested):
Please resolve conflicts with master and squash commits so we have 1 commit for each feature/change.

I hope we can merge after this.

nemesifier (approved):

PR #345, opened 2025-04-01
[deps] Add support for Django 5.x and update channels

### LINKED ISSUE #340: [deps] Add support for Django 5.0 & 5.1 and Python 3.11, 3.12 & 3.13
- Drop support for Django 3.2
- Drop support for Python 3.8

**Depends on** 
- [x] https://github.com/openwisp/openwisp-utils/issues/439
- [x] https://github.com/openwisp/openwisp-users/issues/432

pandafy:
## Checklist

**Blockers**

- [x] https://github.com/openwisp/openwisp-utils/pull/460

Closes #340


requirements-test.txt
```diff
@@ -1,7 +1,4 @@
-openwisp-utils[qa,selenium] @ https://github.com/openwisp/openwisp-utils/tarball/1.2
+openwisp-utils[qa,selenium,channels, channels-test] @ https://github.com/openwisp/openwisp-utils/tarball/issues/388-channels
```
pandafy, line 1: TODO: Update this once PR in openwisp-utils is merged.

nemesifier, line 1:
```suggestion
openwisp-utils[qa,selenium,channels,channels-test] @ https://github.com/openwisp/openwisp-utils/tarball/issues/388-channels
```

openwisp_notifications/types.py
```diff
@@ -111,3 +111,10 @@ def _unregister_notification_choice(notification_type):
             NOTIFICATION_CHOICES.pop(index)
             return
     raise ImproperlyConfigured(f'No such Notification Choice {notification_type}')
+
+
+def get_notification_choices():
```
nemesifier, line 116: Are you sure this is necessary?

I don't see the django docs saying that passing a callable is necessary for [choice fields](https://docs.djangoproject.com/en/5.2/ref/models/fields/#choices). Am I missing something?

setup.py
```diff
@@ -64,6 +64,6 @@ def get_install_requires():
         'License :: OSI Approved :: GNU General Public License v3 (GPLv3)',
         'Operating System :: OS Independent',
         'Framework :: Django',
-        'Programming Language :: Python :: 3.8',
+        'Programming Language :: Python :: 3',
```
nemesifier, line 67: Can you update this consistently with the other modules please?

nemesifier (commented):
A few questions below.

setup.py
pandafy: We don't mention the Python version in any OpenWISP module. I checked the following 

- https://github.com/openwisp/openwisp-utils/blob/master/setup.py
- https://github.com/openwisp/openwisp-users/blob/master/setup.py
- https://github.com/openwisp/openwisp-controller/blob/master/setup.py
- https://github.com/openwisp/openwisp-ipam/blob/master/setup.py

openwisp_notifications/types.py
pandafy: I have added the comment in the code to explain the implementation details of Django 

From the docs: 

> Passing a callable for choices can be particularly handy when, for example, the choices are:
> 
> - the result of I/O-bound operations (which could potentially be cached), such as querying a table in the same or an external database, or accessing the choices from a static file.
> 
> - a list that is mostly stable but could vary from time to time or from project to project. Examples in this category are using third-party apps that provide a well-known inventory of values, such as currencies, countries, languages, time zones, etc.

We fall in the second point. The `NOTIFICATION_CHOICES` changes when registering/unregistering notification types.

nemesifier: Good, point I was forgetting this important detail for a moment.

openwisp_notifications/types.py
```diff
@@ -111,3 +111,10 @@ def _unregister_notification_choice(notification_type):
...
+    """
+    Returns the list of notification choices.
```
nemesifier, line 118:
```suggestion
    Returns the list of notification choices, which may
    be dyanmically changed at runtime by other openwisp
    modules which register new notification types.
```

setup.py
nemesifier: Ok great, maybe the other modules I saw were libraries and we use more classifiers there.

nemesifier (approved):

PR #354, opened 2025-05-14
[fix] Notification preferences API returns fallback value if None #351

### LINKED ISSUE #351: [bug] Notification preferences page shows all notifications as disabled
**Describe the bug**

`NotificationSetting.web` and `Notification.email` fields can have three states: 
- `True` - notification is enabled
- `False` - notification is disabled
- `None` - fallback to the value defined in notification type 

After upgrading the instance to the latest master, we see all preferences as disabled in the UI. 

This happens because the changes made in #344 does not handle `None` values in the field. 

**Steps to replicate**

1. Install OpenWISP Notifications from `1.1` branch. 
2. Run migrations, create a user and ensure the user has NotificationSettings 

Image

3. Upgrade the notifications module to the master branch 

**Expected outcome**

All notification preferences appear enabled. 

**Actual outcome**

All notification preferences appear disabled 

Image

pandafy:
## Checklist

Fixes #351

## Description of Changes

The notification prefences API would return
`NotificationSetting.web_notification` for the `web` field and `NotificationSetting.email_notification` for the `email` field.

These methods returns the fallback value from the notification type configuration, when the database contains `None`.

### DIFF
--- a/openwisp_notifications/api/serializers.py
+++ b/openwisp_notifications/api/serializers.py
@@ -83,6 +83,12 @@ class NotificationSettingSerializer(serializers.ModelSerializer):
         exclude = ['user']
         read_only_fields = ['organization', 'type']
 
+    def to_representation(self, instance):
+        data = super().to_representation(instance)
+        data['web'] = instance.web_notification
+        data['email'] = instance.email_notification
+        return data
+
 
 class IgnoreObjectNotificationSerializer(serializers.ModelSerializer):
     class Meta:
--- a/openwisp_notifications/api/serializers.py
+++ b/openwisp_notifications/api/serializers.py
@@ -6,6 +6,7 @@ from rest_framework.exceptions import NotFound
 
 from openwisp_notifications.exceptions import NotificationRenderException
 from openwisp_notifications.swapper import load_model
+from openwisp_utils.api.serializers import ValidatedModelSerializer
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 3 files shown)


openwisp_notifications/tests/test_selenium.py
```diff
@@ -126,45 +126,71 @@ def test_email_unsubscribe_page(self):
...
+                '.global-setting-dropdown[data-web-state]'
+                ' .global-setting-dropdown-menu button:last-child',
+            ).click()
+            self.find_element(By.CSS_SELECTOR, '#confirmation-modal #confirm').click()
+
+            all_checkboxes = self.find_elements(
+                By.CSS_SELECTOR, 'input[type="checkbox"]', wait_for='presence'
+            )
+            for checkbox in all_checkboxes:
+                self.assertFalse(checkbox.is_selected())
```
nemesifier, line 172: Can we take advantage to ensure the values in the database matches what we expect?

nemesifier (changes requested):
It works, thanks! Can we make the automated test stricter by checking the values in the DB match our expectations after making changes to the UI?

openwisp_notifications/api/serializers.py
```diff
@@ -72,7 +73,7 @@ class Meta(NotificationSerializer.Meta):
         list_serializer_class = CustomListSerializer
 
 
-class NotificationSettingSerializer(serializers.ModelSerializer):
+class NotificationSettingSerializer(ValidatedModelSerializer):
```
pandafy, line 76: The serializer was not using `ValidatedModelSerializer` earlier. Thus, it was not calling the model's full_clean method. Due to this, the database was not storing `None` value. 

Thanks for catching this.

PR #356, opened 2025-05-16
[feature] Suppress email verification prompt when email notifications a…

### LINKED ISSUE #343: [change] Check if user has notification enabled before prompting to verify email address
Follow up of #281 / #336, to be done once we merge the gsoc24-rebased branch in master.

When a user logs in, if he doesn't have any verified email address we say that notifications are enabled and prompt to send the verification email again.

However, we aren't checking if the user really has any email notification enabled.
I think it would make sense to check this as not all users will care about email notifications and prompting them to verify their email address may be annoying for some.

I think we should:

- Provide a method which checks whether the user has any email notification enabled
- Use this method before checking if the user has any verified email in `check_email_verification`.

**Blockers**

- [ ] https://github.com/openwisp/openwisp-notifications/pull/344

stktyagi:
…re disabled #343

Updated the check_email_verification function to check the email boolean value before sending email verification notification. Tested by creating 2 separate users with opposite boolean values and if email=True (email notification enabled) then only user is prompted with notification, else restrained.

Fixes #343

## Checklist

Closes #343.

## Description of Changes

As mentioned in the issue, the check_email_verification function was updated to ensure that only is email boolean value is true (email notifications are enabled) then only user gets prompted for verification.

### DIFF
--- a/openwisp_notifications/handlers.py
+++ b/openwisp_notifications/handlers.py
@@ -400,6 +400,14 @@ def check_email_verification(sender, user, request, **kwargs):
     # abort if this is not an admin login
     if not user.is_staff or not request.path.startswith(admin_path):
         return
+
+    email_notifs_enabled = NotificationSetting.objects.filter(
+        user=user, email=True
+    ).exists()
+    if not email_notifs_enabled:
+        # if no email notifications enabled, don't prompt
+        return
+
     has_verified_email = EmailAddress.objects.filter(user=user, verified=True).exists()
     # abort if user already has a verified email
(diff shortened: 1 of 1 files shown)


nemesifier (changes requested):
Good progress @stktyagi! :+1: 

A test is missing for this, please add one here:
https://github.com/openwisp/openwisp-notifications/blob/master/openwisp_notifications/tests/test_views.py
Look at existing tests for reference. The new test should be similar but we must disable all email notifications before triggering the login and verify no messages were created.

PR #357, opened 2025-05-16
[chores] Avoid email notification warning if email notifications are disabled

### LINKED ISSUE #343: [change] Check if user has notification enabled before prompting to verify email address

stktyagi:
… are disabled #343

Updated the check_email_verification function to check the email boolean value before sending email verification notification. Tested by creating 2 separate users with opposite boolean values and if email=True (email notification enabled) then only user is prompted with notification, else restrained.

Fixes #343

## Checklist

Closes #343.

## Description of Changes

As mentioned in the issue, the check_email_verification function was updated to ensure that only is email boolean value is true (email notifications are enabled) then only user gets prompted for verification. Updated towards class method for better readability. Tests were written as mentioned and manual checking was done by simulating 2 users with different email boolean values.

### DIFF
--- a/openwisp_notifications/handlers.py
+++ b/openwisp_notifications/handlers.py
@@ -394,11 +394,19 @@ def update_notification_cache(sender, instance, **kwargs):
     transaction.on_commit(invalidate_cache)
 
 
+class NotificationUtils:
+    @classmethod
+    def email_notifications_enabled(cls, user):
+        return NotificationSetting.objects.filter(user=user, email=True).exists()
+
+
 @receiver(user_logged_in)
 def check_email_verification(sender, user, request, **kwargs):
     admin_path = reverse('admin:index')
     # abort if this is not an admin login
     if not user.is_staff or not request.path.startswith(admin_path):
+        return
+    if not NotificationUtils.email_notifications_enabled(user):
         return
     has_verified_email = EmailAddress.objects.filter(user=user, verified=True).exists()
     # abort if user already has a verified email
--- a/openwisp_notifications/handlers.py
+++ b/openwisp_notifications/handlers.py
@@ -394,19 +394,13 @@ def update_notification_cache(sender, instance, **kwargs):
     transaction.on_commit(invalidate_cache)
 
 
-class NotificationUtils:
-    @classmethod
-    def email_notifications_enabled(cls, user):
-        return NotificationSetting.objects.filter(user=user, email=True).exists()
-
-
 @receiver(user_logged_in)
 def check_email_verification(sender, user, request, **kwargs):
     admin_path = reverse('admin:index')
     # abort if this is not an admin login
     if not user.is_staff or not request.path.startswith(admin_path):
         return
-    if not NotificationUtils.email_notifications_enabled(user):
+    if not NotificationSetting.email_notifications_enabled(user):
         return
     has_verified_email = EmailAddress.objects.filter(user=user, verified=True).exists()
     # abort if user already has a verified email
--- a/openwisp_notifications/tests/test_views.py
+++ b/openwisp_notifications/tests/test_views.py
@@ -8,7 +8,7 @@ from django.test import TestCase
 from django.test.client import Client
 from django.urls import reverse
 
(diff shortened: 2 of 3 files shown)


stktyagi:
hey @nemesifier,
Apologies but I ran into some issue in my environment so, I've added updates as you mentioned in the review in this PR.

openwisp_notifications/handlers.py
```diff
@@ -394,12 +394,20 @@ def invalidate_cache():
     transaction.on_commit(invalidate_cache)
 
 
+class NotificationUtils:
+    @classmethod
+    def email_notifications_enabled(cls, user):
+        return NotificationSetting.objects.filter(user=user, email=True).exists()
```
nemesifier, line 400: Please move this method to the abstract `NotificationSetting` model, there's no point in having this class here on its own.

Add a short docstring to the method, eg:

```python
"""Returns ``True`` if ``user`` has at least one email notification setting enabled.""""
```

openwisp_notifications/handlers.py
```diff
@@ -394,12 +394,20 @@ def invalidate_cache():
...
+
+
 @receiver(user_logged_in)
 def check_email_verification(sender, user, request, **kwargs):
     admin_path = reverse('admin:index')
     # abort if this is not an admin login
     if not user.is_staff or not request.path.startswith(admin_path):
         return
+    if not NotificationUtils.email_notifications_enabled(user):
```
nemesifier, line 409:
```suggestion
    if not NotificationSetting.email_notifications_enabled(user):
```

openwisp_notifications/tests/test_views.py
```diff
@@ -136,3 +145,20 @@ def test_warning_on_admin_login(self):
         self.assertIn('verify your email address', message)
         expected_url = reverse('notifications:resend_verification_email')
         self.assertIn(expected_url, message)
+
+    def test_warning_disabled(self):
```
nemesifier, line 149: Let's make the test method name more obvious, eg:
```suggestion
    def test_email_notifications_disabled_no_warning(self):
```

openwisp_notifications/tests/test_views.py
```diff
@@ -116,6 +120,11 @@ def setUp(self):
         )
 
     def test_warning_on_admin_login(self):
+        NotificationSetting.objects.update_or_create(
+            user=self.user,
+            type='default',
+            defaults={'email': True, 'web': True},
+        )
```
nemesifier, line 127: I assume future tests in this test class will be based on this case, therefore we could move this to the `setUp` method to avoid duplicating this logic in new tests, which is highly likely if new contributors add new tests in the future and we may miss the duplication (which adds maintenance burden in the long run).

nemesifier (changes requested):
Great progress @stktyagi, see my comments below.

openwisp_notifications/tests/test_views.py
stktyagi: makes sense, I'll add it as a centralized shared object.

stktyagi:
thank you for the review,
the above mentioned and the following changes have been made:-

1.)unused swapper import was failing qa tests, so we got rid of that.

2.)Replaced implicit checks of email notification status with a direct call to the email_notifications_enabled class method.

nemesifier (commented):
Looks good, I am going to tweak a small detail (see below).

Please remember to follow the [commit message guidelines](https://openwisp.io/docs/dev/developer/contributing.html).

nemesifier (approved):

PR #367, opened 2025-06-14
[fix] Hide notification preferences for disabled orgs #366

### LINKED ISSUE #366: [bug] Notification preferences for disabled organizations should be hidden
**Describe the bug**
Notification preferences for disabled organizations are visible.

**Steps To Reproduce**
- Create a new org
- Disable it
- Go to the notification preferences page

**Expected behavior**
The notification preferences for the disabled org are not shown.

stktyagi:
Filtered out notification settings linked to disabled organizations by modifying the queryset. Added a test to ensure preferences from inactive organizations are excluded from user settings.

Fixes #366

## Checklist

- N/A I have updated the documentation.

Closes #366

## Description of Changes

Updated the queryset to exclude notification settings from inactive organizations. Implemented a test to verify that user preferences tied to disabled organizations are not returned.

### DIFF
--- a/openwisp_notifications/tests/test_notification_setting.py
+++ b/openwisp_notifications/tests/test_notification_setting.py
@@ -1,6 +1,7 @@
 from unittest.mock import patch
 
 from django.core.exceptions import ValidationError
+from django.db.models import Q
 from django.db.models.signals import post_save
 from django.test import TransactionTestCase
 
@@ -396,3 +397,18 @@ class TestNotificationSetting(TestOrganizationMixin, TransactionTestCase):
             with self.assertRaises(ValidationError):
                 global_setting.full_clean()
                 global_setting.save()
+
+    def test_notification_setting_inactive_organizations(self):
+        user = self._get_user()
+        active_org = Organization.objects.create(name="active-org", is_active=True)
+        inactive_org = Organization.objects.create(name="inactive-org", is_active=False)
+        NotificationSetting.objects.create(user=user, organization=None)
+        NotificationSetting.objects.create(user=user, organization=active_org)
+        NotificationSetting.objects.create(user=user, organization=inactive_org)
+        qs = NotificationSetting.objects.filter(user=user).filter(
+            Q(organization__isnull=True) | Q(organization__is_active=True)
+        )
+        self.assertEqual(qs.count(), 2)
+        for setting in qs:
+            if setting.organization is not None:
+                self.assertTrue(setting.organization.is_active)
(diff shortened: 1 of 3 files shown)


stktyagi:
@nemesifier I feel like something better can be done for the test referring this change. Any thoughts?

openwisp_notifications/tests/test_notification_setting.py
```diff
@@ -396,3 +397,18 @@ def test_validate_global_notification_setting(self):
...
+        NotificationSetting.objects.create(user=user, organization=None)
+        NotificationSetting.objects.create(user=user, organization=active_org)
+        NotificationSetting.objects.create(user=user, organization=inactive_org)
+        qs = NotificationSetting.objects.filter(user=user).filter(
+            Q(organization__isnull=True) | Q(organization__is_active=True)
+        )
+        self.assertEqual(qs.count(), 2)
+        for setting in qs:
+            if setting.organization is not None:
+                self.assertTrue(setting.organization.is_active)
```
nemesifier, line 414: what's this test for?

nemesifier (changes requested):
Great progress @stktyagi!

I think we need a different test here, one that tests the preferences API and checks the result to ensure the disabled organizations are not visible, the test should fail without the patch in `api/views.py`

openwisp_notifications/tests/test_notification_setting.py
stktyagi: This test as I mentioned has a limitation that it verifies query logic at the model level, not at the API level, so basically testing patch logic would be a poor test. I have update this with a better implementation than this.

openwisp_notifications/tests/test_notification_setting.py
```diff
@@ -396,3 +397,36 @@ def test_validate_global_notification_setting(self):
             with self.assertRaises(ValidationError):
                 global_setting.full_clean()
                 global_setting.save()
+
+    @mock_notification_types
+    def test_preferences_api_excludes_disabled_organizations(self):
+        def create_validated_instance(Model, **data):
```
nemesifier, line 403: There's not point in creating this function here as it's only used once.

openwisp_notifications/tests/test_notification_setting.py
```diff
@@ -396,3 +397,36 @@ def test_validate_global_notification_setting(self):
             with self.assertRaises(ValidationError):
                 global_setting.full_clean()
                 global_setting.save()
+
+    @mock_notification_types
+    def test_preferences_api_excludes_disabled_organizations(self):
```
nemesifier, line 402: This test is effective but should be placed in `test_api.py` for consistency with the existing tests for this view, which are placed there.

nemesifier (changes requested):
Good progress @stktyagi, I believe after addressing the comments below we can merge this.

PR #373, opened 2025-06-25
[change] Improved batch email logic #371

### LINKED ISSUE #371: [feature] Batch notification emails: exclude read notifications & clarify batching logic
**Is your feature request related to a problem? Please describe.**
Currently, batch notification emails may include notifications that have already been read from the web UI.  This can result in redundant or unnecessary emails.

**Describe the solution you'd like**
- Do not include notifications which have already been read in the batch notification email and clarify this detail in the wording used, you have X unread notifications.
- Continue batching notifications after a batch summary email is sent, instead of reverting to individual emails for subsequent notifications.

**Thinking out loud**
What happens if a user always stays on top of their notifications and marks them as read quickly? How will the code determine when to stop batching and return to sending individual notification emails? 

If all notifications in the batch have been read then we don't need to send the summary and we can rever to single emails.

pandafy:
## Checklist

Closes #371

## Description of Changes

- Don't include read notifications in batch email.


openwisp_notifications/handlers.py
```diff
@@ -175,41 +178,21 @@ def send_email_notification(sender, instance, created, **kwargs):
...
-
+    cache_key = f"email_batch_{instance.recipient.id}"
     cache_data = cache.get(
         cache_key,
         {
             "last_email_sent_time": None,
             "batch_scheduled": False,
             "pks": [],
             "start_time": None,
-            "email_id": instance.recipient.email,
```
nemesifier, line 212: Why is this being removed?

nemesifier (commented):
I don't see changes that implement the following:

- Continue batching after a batch is sent
- If there's no unread notifications to be sent in the batch, go back to sending single notifications

It would be good to have a test for each. Were tests already present for these?

openwisp_notifications/handlers.py
pandafy: We already do a lookup for `User` (instance.recipient) in the celery task. Thus, we don't need to store the info in the cache.

nemesifier (commented):
Please update the docs: there's a paragraph which says "Key aspects of the batch email notification feature include", we should add these 2 important points:

- When email batches are sent, only unread notifications are included (question: do we do this check when it's time to actually send the batch, right?) 
- Clarify how the system goes back to sending individual emails. We are changing this logic so that once batches are sent, the system keeps batching right?

pandafy:
> * When email batches are sent, only unread notifications are included (question: do we do this check when it's time to actually send the batch, right?)

Yes, we check this while preparing the batch email. 

> * Clarify how the system goes back to sending individual emails. We are changing this logic so that once batches are sent, the system keeps batching right?

Correct.

nemesifier (approved):

PR #375, opened 2025-06-27
[change] Disallow changing preferences for generic messages #359

### LINKED ISSUE #359: [change] Disallow changing preferences for generic messages
It may not make sense to allow users to configure preferences for **generic messages**.

- Receiving an email for these may be overkill
- At the same time, disabling them entirely doesn't make sense either as this notification type is designed to inform the user about errors that are happening in the background

Instead of exposing configuration options, we could treat these as always-on informational messages and exclude them from the user notification preferences UI.

pandafy:
## Checklist

Closes #359

### DIFF
--- a/openwisp_notifications/settings.py
+++ b/openwisp_notifications/settings.py
@@ -9,6 +9,7 @@ CONFIG_DEFAULTS = {
     "NUM_TO_FETCH": 10,
     "CACHE_TIMEOUT": 2,
 }
+DISALLOW_PREFERENCES_CHANGE_TYPE = ["generic_message"]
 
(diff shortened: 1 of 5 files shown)


openwisp_notifications/settings.py
```diff
@@ -9,6 +9,7 @@
     "NUM_TO_FETCH": 10,
     "CACHE_TIMEOUT": 2,
 }
+DISALLOW_PREFERENCES_CHANGE_TYPE = ["generic_message"]
```
pandafy, line 12: I have not made this setting configurable. We can do so if required in future.

PR #379, opened 2025-07-02
[change] Remove unnecessary TEMPLATE DIRS in settings.py

asmodehn:
## Checklist

- [X] I have read the [OpenWISP Contributing Guidelines](http://openwisp.io/docs/developer/contributing.html).
- [X] I have manually tested the changes proposed in this pull request.

## Description of Changes

It seems the setting `TEMPLATE[0]["DIRS"]` is unnecessary, since we are using loaders to resolve templates.

Also looking at [the docs] (https://docs.djangoproject.com/en/5.2/howto/overriding-templates/#overriding-from-the-project-s-templates-directory) it seems this setting does not apply to the current situation (django project template override vs extended app template)

And finally, imho, it makes it even harder to understand  what one needs to do to write an `openwisp-notifications-extended` app with templates that work as expected...

I removed that setting on my machine and tests still pass... But I m not used to django and might have misunderstood something ? or it might be something needed for an old version of django only ?

### DIFF
--- a/tests/openwisp2/settings.py
+++ b/tests/openwisp2/settings.py
@@ -207,14 +207,6 @@ if os.environ.get("SAMPLE_APP", False):
     OPENWISP_NOTIFICATIONS_NOTIFICATIONSETTING_MODEL = (
         "sample_notifications.NotificationSetting"
     )
-    TEMPLATES[0]["DIRS"] = [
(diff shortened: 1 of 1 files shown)


asmodehn:
Ok it seems these lines are necessary to workaround a problem in template loading, when extending openwisp-notifications. See #380

PR #400, opened 2025-08-14
[change] Batch email sending: restore standard behavior if only 1 notification is sent

### LINKED ISSUE #396: [change] Batch email sending: restore standard behavior if only 1 notification is sent
Single notifications are faster to read when received via email and can be acknowledged and deleted without opening them.

If there's only 1 email in the batch, can we send it singularly instead of using the batch template?
When this happens we can restore the standard sending as well so the next notification will also be sent as a single notification.

We'll need to update the docs to add a new section clarifying how the standard behavior is restored for future reference.

It would work more or less as follows:

```
if the batch contains only 1 notification:
    send the notification singularly instead of using the batch email template
    clear the batch
    restore sending single notifications
```

The batching will kick in again if there's a burst of notifications, I don't think we need to make any change for this as it should be already implemented.

Eeshu-Yadav:
## Checklist

Closes #396.

## Description of Changes

**Solution**: 
- Detect when a batch contains only 1 notification
- Send it individually using `notification.send_email()` instead of the batch template
- Clear the batch state by setting an old timestamp to restore standard behavior for subsequent notifications
- Enhanced error handling with proper None-checking for timestamp comparisons

**Changes**:
- Modified `send_batched_email_notifications` task to handle single notification batches
- Enhanced email handler with improved None-checking
- Fixed subject line generation for consistent notification counts
- Added comprehensive test coverage for the new behavior

### DIFF
--- a/openwisp_notifications/handlers.py
+++ b/openwisp_notifications/handlers.py
@@ -194,6 +194,7 @@ def send_email_notification(sender, instance, created, **kwargs):
     #    no batch is scheduled
     if not last_email_sent_time or (
         not batch_start_time
+        and last_email_sent_time  # Check that last_email_sent_time is not None
         and (
             # More than EMAIL_BATCH_INTERVAL seconds have passed since the last email was sent
             last_email_sent_time
--- a/openwisp_notifications/tasks.py
+++ b/openwisp_notifications/tasks.py
@@ -8,6 +8,7 @@ from django.db.models import Q
 from django.db.utils import OperationalError
 from django.utils import timezone
 
+from openwisp_notifications import settings as app_settings
 from openwisp_notifications import types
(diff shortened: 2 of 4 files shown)


nemesifier (commented):
It looks like it does what we need, I couldn't get to test this yet, I ask @pandafy to step in when he can here.

pandafy (changes requested):
I performed a manual test of this PR and the changes does not achieve the desired goal. 

Based on the recent discussion in the issue, we want to render the single notification using the notification template. The email should look similar to this 

image

The email subject and the title in the HTML email should be `notification.email_subject`.

Right now, I see the email subject and HTML title is still `1 unread notification` 

Screenshot from 2025-08-19 17-20-43

PR #417, opened 2025-10-04
[fix] Disconnect websocket on page reload to prevent missed notificat…

### LINKED ISSUE #264: [ui:bug] Notification Toast Container Disappears
I encountered a bug while implementing background API tasks for the [ZeroTier VPN backend in openwisp-controller](https://github.com/openwisp/openwisp-controller/pull/778#pullrequestreview-1513189523).

When a background API task encounters an unrecoverable failure or recovers from any API failures, it generates an error or recovery type notification, respectively. However, if this notification is generated right after saving or updating the VPN object, the notification toast gets automatically dismissed due to the already-present Django model object's success message.

api_task_notification_bug

I tried to address this issue by adding some delay to the notification, but this solution may not be optimal. We need to find a better and more reliable solution to prevent the notification toast from disappearing prematurely.

Adityashandilya555:
When users click "Save" or "Save and Continue", the page reloads and closes the websocket connection. If a background task completes during this reload, the notification is lost because there's no active websocket connection to receive it.

Fixes #264

## Checklist

Closes #264 .

## Description of Changes
Added websocket disconnection in the `beforeunload` event handler. This
   ensures the server knows the client is unavailable and will queue notifications.
  When the page reloads and the websocket reconnects, any queued notifications will be
  delivered properly

## Screenshot

Please include any relevant screenshots.

### DIFF
--- a/openwisp_notifications/static/openwisp-notifications/js/notifications.js
+++ b/openwisp_notifications/static/openwisp-notifications/js/notifications.js
@@ -32,6 +32,10 @@ const owNotificationWindow = {
     // closing the window
     $(window).on("beforeunload", function () {
       owNotificationWindow.remove();
+      // Disconnect websocket to prevent missed notifications during page reload
+      if (typeof notificationSocket !== "undefined") {
+        notificationSocket.close();
+      }
     });
     // Get authority to play notification sound when
     // other windows are closed


Adityashandilya555:
i have fixed the qa formatting issue

PR #428, opened 2025-11-26
[fix] Fixed create_notification_settings task overriding user preferences #427

### LINKED ISSUE #427: [bug] Upgrading from 1.1.0 to 1.2.0 resets notification preferences
**Describe the bug**
Upgrading from OpenWISP Notifications 1.1.0 to 1.2.0 resets user notification preferences to `None` in the database.

**Steps To Replicate**

1. Checkout 1.1.0 version of openwisp-notifications and install all the dependencies. Start a celery worker and run the django server.
2. Create a superuser.
3. Open the superuser's change page and disable all the notifications.
4. Stop the django server and the celery worker. Checkout the 1.2.0 version of openwisp-notifications. Install all dependencies and start django server and celery worker.
5. Open the notification preferences page for the user.

**Expected behavior**
The notification preferences should stay as is after an upgrade.

**Actual Result**
The notification preferences get reset to `None` in the database after the upgrade.

**Additional notes**
---
Here's the SQL output from the database on version 1.1.0 after disabling all notifications from django admin:
```sql
SELECT id,type,web,email FROM openwisp_notifications_notificationsetting;
e8c790481c5d43639a66dcd760f9fab0    | default        | 0  | 0
3e1f93f637fa40a48a40f23578bccbb2    | generic_message| 0  | 0
```
Here's the SQL output from the database after upgrading to version 1.2.0:
```
SELECT id,type,web,email FROM openwisp_notifications_notificationsetting;
e8c790481c5d43639a66dcd760f9fab0    |default        |      |
3e1f93f637fa40a48a40f23578bccbb2    |generic_message|      |
05de43baa3d04fadbd17aafde329b330    |               | 1    | 1
```

As you can see the user configured value for the NotificationSetting has been erased.

pandafy:
## Checklist

Fixes #427

Please include any relevant screenshots.

### DIFF
--- a/openwisp_notifications/tasks.py
+++ b/openwisp_notifications/tasks.py
@@ -1,7 +1,6 @@
 import logging
 from datetime import timedelta
 
-import django
 from celery import shared_task
 from django.contrib.auth import get_user_model
 from django.contrib.contenttypes.models import ContentType
@@ -95,32 +94,20 @@ def create_notification_settings(user, organizations, notification_types):
             except ObjectDoesNotExist:
                 email = app_settings.WEB_ENABLED
                 web = app_settings.EMAIL_ENABLED
-            if django.VERSION <= (5, 0):
-                # TODO: Remove this condition when support for Django 4.2 is dropped
-                updated = NotificationSetting.objects.filter(
-                    user=user, type=type, organization=org
-                ).update(deleted=False)
-                if not updated:
-                    # Create notification setting if it does not exist
-                    NotificationSetting.objects.create(
-                        user=user,
-                        type=type,
-                        organization=org,
-                        email=None if email else False,
-                        web=None if web else False,
-                        deleted=False,
-                    )
-            else:
-                NotificationSetting.objects.update_or_create(
-                    create_defaults={
-                        "deleted": False,
-                        "email": None if email else False,
-                        "web": None if web else False,
-                    },
-                    defaults={"deleted": False},
+            # If NotificationSetting already exists, then we ensure it is not marked deleted
+            updated = NotificationSetting.objects.filter(
+                user=user, type=type, organization=org
+            ).update(deleted=False)
+            if not updated:
+                # "updated" will be 0 if no NotificationSetting was found.
+                # In this case, we create a new NotificationSetting.
+                NotificationSetting.objects.create(
                     user=user,
                     type=type,
                     organization=org,
+                    email=None if email else False,
+                    web=None if web else False,
+                    deleted=False,
                 )
 
 


openwisp_notifications/tasks.py
```diff
@@ -94,16 +95,33 @@ def create_notification_settings(user, organizations, notification_types):
...
-            NotificationSetting.objects.update_or_create(
-                defaults={
-                    "deleted": False,
-                    "email": None if email else False,
-                    "web": None if web else False,
-                },
-                user=user,
-                type=type,
-                organization=org,
-            )
```
pandafy: **TL;DR:**  
The post-migrate task uses update_or_create(), which applies defaults during both creation and updates. 

Because the defaults come from the organization-level settings, not the user’s existing preferences, any existing NotificationSetting gets overwritten.

This causes user-configured notification preferences to be unintentionally reset to organization defaults.

---

### Explanation of the Issue

The `post_migrate` handler calls this task to ensure that every newly registered notification type has a corresponding `NotificationSetting` for each user–organization pair.  
The code uses `update_or_create()` with the following defaults:

```python
defaults={
    "deleted": False,
    "email": None if email else False,
    "web": None if web else False,
}
```

The problem is that update_or_create() applies the defaults dictionary not only when creating a new NotificationSetting, but also when updating an existing one.

Organization-level notification settings were introduced in version 1.2.0, and they are enabled by default. When this code runs, the variables `email` and web come from:

```
org_notification_settings = org.notification_settings
email = org_notification_settings.email
web = org_notification_settings.web
```

That means these values represent organization-level defaults, not the user’s existing preference stored in NotificationSetting.

As a result, when update_or_create() encounters an existing record:

- It performs an update,
- And overwrites the user’s configuration with email and web derived from the organization-level defaults,
- Causing previously customized user preferences to be lost.

**In short, this logic unintentionally resets user notification settings whenever new notification types are processed.**

openwisp_notifications/tasks.py
```diff
@@ -94,16 +95,33 @@ def create_notification_settings(user, organizations, notification_types):
...
+                if not updated:
+                    # Create notification setting if it does not exist
+                    NotificationSetting.objects.create(
+                        user=user,
+                        type=type,
+                        organization=org,
+                        email=None if email else False,
+                        web=None if web else False,
+                        deleted=False,
+                    )
```
pandafy: `update()` returns the number of rows updated. 

updated will be 0 only when there were no previous NotificationSetting for the user. In this case, we can safely create new NotificationSetting using organization default.

PR #434, opened 2025-12-08
Fix: Global Notification Disable Does Not Update Organization-Level Settings

### LINKED ISSUE #431: [bug] Disabling already-disabled global notification does not update organization-level settings
**Describe the bug**
Disabling the already disabled global notification setting does not disable Organization-level notification settings. This results in an inconsistent state where the UI reflects all notifications are disabled, but organization-level notifications are still enabled.

**Steps To Reproduce**

1. Create a superuser and navigate to http://127.0.0.1:8000/notifications/preferences/
2. Ensure all notifications for the user are enabled.
3. Disable the global email settings.
4. Step 3 will disable email notifications for all notification types and organizations.
5. Enable the email notifications for the **default** organization.
6. At this point, you will have global email notification disabled, but the email notifications for the **default** organization should be enabled.
7. Try to disable the global email notifications again:
    1. Click on the global email notification setting dropdown
    2. Select "Don't notify by email"
    3. Click on "Confirm" on the popup
8. UI will update reflecting that emails for all organizations are disabled
9. Refresh the page

**Expected behavior**
All email notifications are disabled

**Actual result**
Global email notifications are disabled as expected.
Organization email notifications are still enabled.

https://github.com/user-attachments/assets/2f782d2b-d155-4ec5-8d84-ca18917b10c4

**Proposed solution**
- Do not allow clicking on the selected global notification setting. 
   In order to disable all notifications, the user will first have to enable the global notification and then disable it again. This solution has poor UX. 
- Change the view such that disabling the global notification setting forces disabling all notifications.

AditiSah05:
## Checklist

Closes #431 .

Please [open a new issue](https://github.com/openwisp/openwisp-notifications/issues/new/choose) if there isn't an existing issue yet.

## Description of Changes
The issue occurred because disabling the global email notification setting did not automatically update the organization-level notification settings, resulting in a mismatch between what the UI displayed and the actual backend state.

-To resolve this, the update logic for global notification preferences has been modified so that:

-Whenever the global email notification is disabled,

-All organization-level email notification preferences for that user and channel are also force-disabled,

-Ensuring that no organization-level settings remain enabled when the global setting is off.

This synchronization guarantees that both the UI and backend remain consistent.
The solution ensures that:

1.The UI correctly shows all email notifications as disabled.

2.After refreshing the page, organization-level notifications remain disabled.

3.There is no longer any possibility of having global email notifications turned off while organization-level email notifications are still active.

The attached video demonstrates the corrected behavior, confirming that the system now updates both global and organization-level preferences properly.

https://github.com/user-attachments/assets/28ab64b8-1f0a-4a39-b840-e398df39a78a

### DIFF
--- a/openwisp_notifications/api/views.py
+++ b/openwisp_notifications/api/views.py
@@ -145,6 +145,16 @@ class NotificationSettingListView(BaseNotificationSettingView, ListModelMixin):
 class NotificationSettingView(BaseNotificationSettingView, RetrieveUpdateAPIView):
     lookup_field = "pk"
 
+    def get_object(self):
+        if self.kwargs["pk"] == "null":
+            # Handle global setting (organization=None, type=None)
+            return get_object_or_404(
+                self.get_queryset(),
+                organization__isnull=True,
+                type__isnull=True
+            )
+        return super().get_object()
+
 
 class BaseIgnoreObjectNotificationView(GenericAPIView):
     model = IgnoreObjectNotification
--- a/tests/openwisp2/urls.py
+++ b/tests/openwisp2/urls.py
@@ -2,9 +2,14 @@ import os
 
 from django.contrib import admin
 from django.contrib.staticfiles.urls import staticfiles_urlpatterns
+from django.shortcuts import redirect
 from django.urls import include, path
 
+def home(request):
+    return redirect('admin:index')
+
 urlpatterns = [
+    path("", home, name="home"),
     path("admin/", admin.site.urls),
     path("accounts/", include("allauth.urls")),
     path("api/v1/", include("openwisp_utils.api.urls")),
--- a/create_superuser.py
+++ b/create_superuser.py
@@ -1,0 +1,16 @@
+import os
+import sys
+import django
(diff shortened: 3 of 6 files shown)


tests/openwisp2/urls.py
pandafy: Can you explain why these changes are required?

openwisp_notifications/api/views.py
```diff
@@ -145,6 +145,16 @@ def get(self, request, *args, **kwargs):
...
+    def get_object(self):
+        if self.kwargs["pk"] == "null":
+            # Handle global setting (organization=None, type=None)
+            return get_object_or_404(
+                self.get_queryset(),
+                organization__isnull=True,
+                type__isnull=True
+            )
+        return super().get_object()
+
```
pandafy: Can you explain the logic behind this change? Why do we return an object when pk is None?

The current state of code already handles changing of global notification settings. We only need to handle the corner case where user re-attempts to disable the already disabled global notification setting.

PR #445, opened 2026-03-06
[fix] Exclude deleted notification settings #426

### LINKED ISSUE #426: [bug] User notification preferences page throws error after removing notification type and repopulating preferences
**Describe the bug**
The `populate_notification_preferences` command does not delete the NotifiationSettings for undefined Notification Types, instead it flags them as deleted. 

The REST API endpoint does not exclude the deleted NotificationSetting objects. Thus, opening the notification preferences page raises NotificationRenderException. 

**Steps To Reproduce**

1. Ensure there's a superuser created in your project
2. Add a new notification type in `openwisp_notifications/apps.py` using the [`register_notification_type` method](https://openwisp.io/docs/dev/notifications/developer/utils.html#register-notification-type)
3. Run `./manage.py populate_notification_preferences` command
4. Remove the code which adds the new notification type
5. Run `./manage.py populate_notification_preferences` command which is supposed to delete NotificationSettings for undefined notification types. 
6. Run django server and go to the user's notification preferences page

**Expected behavior**
User notification preferences are displayed.

**Actual result**
No notification preference is displayed. The application shows `Error fetching notification settings. Please try again.`

**Other Notes**

This bug was found while upgrading OpenWISP Controller from 1.1 to 1.2 (the `api_task_error` and `api_task_recovery` notification types were removed in openwisp-controller). Thus, the solution should also be tested manually with OpenWISP controller.

Piyushkhobragade:
The populate_notification_preferences command soft-deletes obsolete
notification settings by setting deleted=True. However, these settings
were not being filtered out by the REST API. 

This caused the frontend to throw a NotificationRenderException when 
attempting to render undefined notification types. This PR explicitly 
excludes these soft-deleted records from the BaseNotificationSettingView 
queryset and the UserOrgNotificationSettingView bulk updates. A test case 
has also been added to prevent regressions.

Fixes #426

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -189,3 +189,4 @@ venv.bak/
 pip-selfcheck.json
 
 # End of https://www.gitignore.io/api/code,venv,django,python
+venv/
--- a/.gitignore
+++ b/.gitignore
@@ -189,4 +189,3 @@ venv.bak/
 pip-selfcheck.json
 
 # End of https://www.gitignore.io/api/code,venv,django,python
-venv/
--- a/openwisp_notifications/api/views.py
+++ b/openwisp_notifications/api/views.py
@@ -130,6 +130,7 @@ class BaseNotificationSettingView(GenericAPIView):
         return NotificationSetting.objects.exclude(
             Q(organization__is_active=False)
             | Q(type__in=app_settings.DISALLOW_PREFERENCES_CHANGE_TYPE)
+            | Q(deleted=True)
         ).filter(user_id=user_id)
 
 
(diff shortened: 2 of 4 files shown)


nemesifier (commented):
Looks good, will test asap, please change the prefix from `[api]` to `[fix]`.

Piyushkhobragade:
> Looks good, will test asap, please change the prefix from `[api]` to `[fix]`.

Thanks so much for the review! I initially wasn't sure if updating the PR title was enough, but I just checked the GitHub Actions logs and saw the commit linter was actively failing the build because of the old [api] prefix.

I went ahead and squashed the commits and amended the message to use the [fix] prefix so the CI pipeline can pass. Let me know if you need anything else!

Piyushkhobragade:
Hi @nemesifier, just a quick bump on this! The API fix and commit prefix are all set.

I noticed the CI is currently failing globally due to an upstream django-filter dependency drift. Let me know if you need anything else from my end, or if we are just waiting on the infrastructure fix before you test locally. Thanks!

nemesifier:
I'm sorry for my late reply, we got caught in an avalanche of PRs. This problem was solved by #460. Thanks for contributing!

PR #446, opened 2026-03-21
[fix] Resync org settings on repeated global email disable #431

### LINKED ISSUE #431: [bug] Disabling already-disabled global notification does not update organization-level settings

hypnoastic:
## What does this PR do?

  This fixes a notification preferences bug where reapplying the global email disable state did not resync the user’s organization-level notification settings.

  ## What changed and why?

  - Preserve explicit global preference updates coming from the preferences API.
  - Resync organization-level notification settings even when global email is set to False again and the stored global value was already False.
  - Add regression coverage for the exact API flow used by the notification preferences page.

  ## UI Verification

https://github.com/user-attachments/assets/4da8f4da-0cae-45d1-bfee-a5ad4583ae9e

  ## Self Review Checklist

  - [x] Kept the fix minimal and limited to the affected notification preferences flow
  - [x] Added regression coverage for the issue
  - [x] Verified focused and broader relevant tests pass
  - [x] Avoided unrelated refactors

  Fixes #431

### DIFF
--- a/openwisp_notifications/api/serializers.py
+++ b/openwisp_notifications/api/serializers.py
@@ -91,6 +91,16 @@ class NotificationSettingSerializer(ValidatedModelSerializer):
         data["email"] = instance.email_notification
         return data
 
+    def update(self, instance, validated_data):
+        explicit_global_fields = {
+            field for field in ["web", "email"] if field in validated_data
+        }
+        if explicit_global_fields and not instance.organization and not instance.type:
+            # The preferences UI can explicitly reapply the same global state.
+            # Preserve that intent so the model can resync organization rows.
+            instance._explicit_global_update_fields = explicit_global_fields
+        return super().update(instance, validated_data)
+
 
 class IgnoreObjectNotificationSerializer(serializers.ModelSerializer):
     class Meta:
--- a/openwisp_notifications/base/models.py
+++ b/openwisp_notifications/base/models.py
@@ -445,15 +445,26 @@ class AbstractNotificationSetting(UUIDModel):
                 raise ValidationError("There can only be one global setting per user.")
 
     def save(self, *args, **kwargs):
+        explicit_global_update_fields = set(
+            getattr(self, "_explicit_global_update_fields", [])
+        )
+        if hasattr(self, "_explicit_global_update_fields"):
+            delattr(self, "_explicit_global_update_fields")
         if not self.web_notification:
             self.email = self.web_notification
         with transaction.atomic():
             if not self.organization and not self.type:
                 try:
-                    previous_state = self.__class__.objects.only("email").get(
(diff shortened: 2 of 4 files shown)


hypnoastic:
Forced Push to keep the commit history clean before review.

hypnoastic:
I have addressed all the review feedback given by CodeRabbit and the PR is ready for a review.

nemesifier (changes requested):
Thanks @hypnoastic, this is a well-targeted fix for #431. The bug is subtle: reapplying the same global disable state did not resync organization-level rows because the previous value already matched. Threading an explicit-intent signal from the serializer to `save()` so it can resync even when the stored value is unchanged is a reasonable approach. The regression test that mirrors the actual preferences API flow is exactly what this needs.

A couple of points:

1. The `_explicit_global_update_fields` instance-attribute handoff is a bit fragile. It works, but passing state from serializer to model via a temporary attribute that `save()` reads and `delattr`s is easy to break later. Please add a short comment at both ends explaining the contract, and make sure any other code path that saves a global `NotificationSetting` (admin, bulk ops) still behaves correctly when the attribute is absent.
2. Please confirm the intended semantics in the test: reapplying global email=False should re-disable the per-organization email rows the user had toggled on. The test should assert that resync explicitly, since that is the behavior change.

Also, this PR currently has merge conflicts. Please rebase on the latest master when you push the updates.

Looks good overall. Add the comments, confirm the resync semantics, and fix the conflicts, and we are good.

PR #450, opened 2026-05-06
[fix] Fixed multiple bugs with notification preferences #448

### LINKED ISSUE #448: [bug] NotificationSettings logic ignores user preferences when creating/updating settings
**Describe the bug**
Multiple logic bugs are present in `NotificationSettings` where user notification preferences are ignored when settings are created or updated. Users' explicit choices (global or org-specific) are overwritten or bypassed, leading to unexpected notification behavior.

- **Bug 1: New Organization Creation Ignores User's Global Preferences**
  - **Details:**
    - When a new organization is created, notification settings for all superusers use hardcoded defaults (`email=True`, `web=True`), disregarding their globally configured preferences. The code in `create_notification_settings` and the `ns_organization_created` task uses org settings or defaults only, never inheriting from the user's global prefs.
    - **Location:**
      - Primary: `tasks.py:83-111` (`create_notification_settings()`)
      - Trigger: `tasks.py:212-224` (`ns_organization_created` task)
    - **Root Cause:**
      - Hardcoded defaults are set (lines 84-86) via `get_or_create` for globals.
      - When creating org-specific settings (lines 104-111), user’s global prefs are ignored in favor of org settings.
  - **Expected Behavior:**
    - New org notification settings should inherit from a user's global notification preferences. If user has disabled email/web globally, org settings must respect this.

- **Bug 2: Org Admin Addition Ignores Global Preferences**
  - **Details:**
    - When adding a user as org admin, notification settings are based just on org’s settings, not on the user's global notification preferences. This is handled via `update_org_user_notificationsetting` which calls the flawed `create_notification_settings`.
    - **Location:**
      - Primary: `tasks.py:175-198` (`update_org_user_notificationsetting` task)
      - Trigger: `handlers.py:298-306` (`organization_user_post_save` signal handler)
    - **Root Cause:**
      - Same as Bug 1; global user preferences are omitted.
  - **Expected Behavior:**
    - Adding an org admin should set notification defaults based on user's global preferences.

- **Bug 3: Org Settings Changes Overwrite User Preferences**
  - **Details:**
    - When an org admin changes their organization's notification settings, all users' organization-notification settings are updated en masse without checking for explicit overrides. Thus, even user-customized settings are overwritten.
    - **Location:**
      - Primary: `base/models.py:564-583` (`AbstractOrganizationNotificationSettings._update_organizationuser_settings()`)
    - **Root Cause:**
      - The update logic applies new org settings to all NotificationSetting objects for that org without checking if users have manually customized their prefs.
  - **Expected Behavior:**
    - Only users who have not personally customized their org notification settings should be updated when the org's default changes.

**Steps To Reproduce**
Steps to reproduce each bug are summarized below:

1. Create a user with global notification settings disabling email and/or web notifications. Create a new org as superuser -> User receives enabled notices anyway.
2. Promote/add a user as org admin. Their notification type is enabled in new org regardless of their global preference.
3. As org admin, update org notification settings. All users’ org notification preferences are overwritten even if previously customized.

**Expected behavior**
- User’s explicit global or org-level notification preferences are never overwritten or ignored by system defaults or org changes.

pandafy:
## Checklist

Fixes #448

## Description of Changes
Fixed multiple bugs in notification preference resolution and sending logic.

Key changes:
- Introduced a three-tier resolution order for notification preferences: **user setting → org setting → type default**. Removed the old behavior where changing org-level settings would cascade down and overwrite user-level preferences stored in `NotificationSetting`.
- Added `normalize_settings()` on `AbstractNotificationSetting` to compute whether a user's setting matches the resolved default (storing `None` when it does, avoiding redundant DB rows).
- Updated `notify_handler` to consult organization-level web settings when determining which users to include for a given notification.
- Reworked `create_notification_settings` task to respect global notification preferences when provisioning settings for new organizations or org-users.
- Optimized the notification settings API view with `select_related` to reduce database queries.

### DIFF
--- a/docs/user/notification-preferences.rst
+++ b/docs/user/notification-preferences.rst
@@ -23,30 +23,55 @@ screenshot below:
 Alternatively, you can also visit ``/notification/preferences/`` to manage
 your settings.
 
+Preference Resolution
+---------------------
+
+Notification preferences are resolved using the following inheritance
+chain:
+
+1. User-specific notification preference
+2. Organization notification preference
+3. Notification type default
+
+A notification setting can therefore either:
+
+- explicitly enable notifications,
+- explicitly disable notifications,
+- or inherit the effective value from organization and notification type
+  defaults.
+
 .. note::
 
     - You can disable notifications globally while still enabling them for
-      specific organizations.
+      specific organizations or notification types.
     - Notification settings are now linked: disabling web notifications
       will automatically disable email notifications, and enabling email
       notifications will automatically enable web notifications.
+    - Notification settings inherit organization and notification type
+      defaults whenever possible instead of storing redundant explicit
+      values.
     - Deleting notification settings is no longer possible via the web
       interface (please use the REST API if removal is needed).
 
-Notification settings are automatically generated for all notification
-types and organizations for every user. Superusers have the ability to
-manage notification settings for all users, including adding or deleting
-them. Meanwhile, staff users can modify their preferred notification
-delivery methods, choosing between receiving notifications via web, email,
-or both. Additionally, users have the option to disable notifications
-entirely by turning off both web and email notification settings.
+Notification settings are automatically generated for notification types
+and organizations associated with a user. Effective notification behavior
+is resolved dynamically using inherited defaults. Superusers have the
+ability to manage notification settings for all users, including adding or
+deleting them. Meanwhile, staff users can modify their preferred
+notification delivery methods, choosing between receiving notifications
+via web, email, or both. Additionally, users have the option to disable
+notifications entirely by turning off both web and email notification
+settings.
 
 .. note::
 
-    If a user has not configured their preferences for email or web
-    notifications for a specific notification type, the system will
-    default to using the ``email_notification`` or ``web_notification``
-    option defined for that notification type.
+    If a user has not explicitly configured a notification preference, the
+    system resolves the effective value using the following order:
+
+    1. User notification preference
+    2. Organization notification preference
+    3. Notification type default (``email_notification`` /
+       ``web_notification``)
 
 Organization Settings
 ---------------------
@@ -72,9 +97,9 @@ notification settings from the **Organization Admin**:
 Key points:
 
 - New users inherit these default settings automatically.
-- Changes apply to all existing users. For example, disabling email
-  notifications at the organization level disables them for all users in
-  the organization.
+- Organization settings act as defaults for all users in the organization.
+- Users inheriting organization defaults will automatically observe
+  organization-level changes.
 - Users can override these defaults by customizing their own notification
   preferences.
 - Organization settings override global defaults defined in Django
--- a/docs/user/notification-preferences.rst
+++ b/docs/user/notification-preferences.rst
@@ -47,9 +47,6 @@ A notification setting can therefore either:
     - Notification settings are now linked: disabling web notifications
       will automatically disable email notifications, and enabling email
       notifications will automatically enable web notifications.
-    - Notification settings inherit organization and notification type
-      defaults whenever possible instead of storing redundant explicit
-      values.
     - Deleting notification settings is no longer possible via the web
       interface (please use the REST API if removal is needed).
 
--- a/openwisp_notifications/base/models.py
+++ b/openwisp_notifications/base/models.py
@@ -453,13 +453,17 @@ class AbstractNotificationSetting(UUIDModel):
             if self.organization and hasattr(
                 self.organization, "notification_settings"
             ):
-                should_enable_email = (
-                    should_enable_email
-                    and self.organization.notification_settings.email
-                )
-                should_enabled_web = (
-                    should_enabled_web and self.organization.notification_settings.web
-                )
+                if self.organization.notification_settings.web is not None:
+                    should_enabled_web = (
+                        should_enabled_web
+                        and self.organization.notification_settings.web
+                    )
+                if self.organization.notification_settings.email is not None:
+                    should_enable_email = (
+                        should_enable_email
+                        and self.organization.notification_settings.email
+                    )
+
             if self.email == should_enable_email:
                 self.email = None
             if self.web == should_enabled_web:
--- a/openwisp_notifications/base/models.py
+++ b/openwisp_notifications/base/models.py
@@ -557,27 +557,4 @@ class AbstractOrganizationNotificationSettings(models.Model):
     def save(self, *args, **kwargs):
         if not self.web:
             self.email = False
-        if not self._state.adding:
-            self._update_organizationuser_settings()
         return super().save(*args, **kwargs)
-
-    def _update_organizationuser_settings(self):
-        try:
-            db_instance = self.__class__.objects.only("web", "email").get(
-                organization_id=self.organization_id
-            )
-        except self.__class__.DoesNotExist:
-            return
-        update_fields = {}
-        for field in ["web", "email"]:
-            if getattr(self, field) != getattr(db_instance, field):
-                update_fields[field] = getattr(self, field)
-        if update_fields:
-            NotificationSetting = swapper.load_model(
-                "openwisp_notifications", "NotificationSetting"
-            )
-            transaction.on_commit(
-                lambda: NotificationSetting.objects.filter(
-                    organization_id=self.organization_id
-                ).update(**update_fields)
-            )
--- a/openwisp_notifications/base/models.py
+++ b/openwisp_notifications/base/models.py
@@ -446,27 +446,22 @@ class AbstractNotificationSetting(UUIDModel):
 
     def normalize_settings(self):
         if self.web_notification is False:
-            self.email = self.web_notification
+            self.email = False
         if self.organization and self.type:
             should_enable_email = self.type_config["email_notification"]
-            should_enabled_web = self.type_config["web_notification"]
-            if self.organization and hasattr(
-                self.organization, "notification_settings"
-            ):
-                if self.organization.notification_settings.web is not None:
-                    should_enabled_web = (
-                        should_enabled_web
-                        and self.organization.notification_settings.web
-                    )
-                if self.organization.notification_settings.email is not None:
-                    should_enable_email = (
-                        should_enable_email
-                        and self.organization.notification_settings.email
-                    )
+            should_enable_web = self.type_config["web_notification"]
+            if hasattr(self.organization, "notification_settings"):
+                should_enable_web = (
+                    should_enable_web and self.organization.notification_settings.web
+                )
+                should_enable_email = (
+                    should_enable_email
+                    and self.organization.notification_settings.email
+                )
 
             if self.email == should_enable_email:
                 self.email = None
-            if self.web == should_enabled_web:
+            if self.web == should_enable_web:
                 self.web = None
 
     def save(self, *args, **kwargs):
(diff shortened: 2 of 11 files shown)


openwisp_notifications/tests/test_selenium.py
```diff
@@ -270,6 +272,112 @@ def test_empty_notification_preference_page(self):
...
+        Organization.objects.all().delete()
+        self.login()
+        org = self._get_org()
+        setting = NotificationSetting.objects.get(
+            user=self.admin,
+            organization=org,
+            type="default",
+            web=None,
+            email=None,
+        )
```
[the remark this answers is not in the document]
pandafy: > * Configures `CELERY_TASK_ALWAYS_EAGER=True` for test execution

@coderabbitai this is already present in all OpenWISP test projects.

openwisp_notifications/handlers.py
```diff
@@ -81,14 +88,16 @@ def notify_handler(**kwargs):
...
+        # Users with web=None are included only if the fallback
+        # chain can still resolve to True.
         if notification_type:
-            # Create notification for users who have opted for receiving notifications.
-            # For users who have not configured web_notifications,
-            # use default from notification type
             web_notification = Q(notificationsetting__web=True)
-            if notification_template["web_notification"]:
+            if org_web is True and notification_template["web_notification"]:
                 web_notification |= Q(notificationsetting__web=None)
```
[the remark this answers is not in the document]
pandafy: @coderabbitai does this bug still exist?

pandafy:
@coderabbitai full review!

openwisp_notifications/base/models.py
```diff
@@ -490,13 +509,29 @@ def type_config(self):
...
+                email_enabled and self.organization.notification_settings.email
+            )
+        return email_enabled
 
     @property
     def web_notification(self):
         if self.web is not None:
             return self.web
-        return self.type_config.get("web_notification")
+        web_enabled = self.type_config.get("web_notification")
```
nemesifier, line 571: a docstring summarizing the logic here would be good

openwisp_notifications/base/models.py
```diff
@@ -490,13 +509,29 @@ def type_config(self):
     def email_notification(self):
         if self.email is not None:
             return self.email
-        return self.type_config.get("email_notification")
+        email_enabled = self.type_config.get("email_notification")
```
nemesifier, line 545: same here

openwisp_notifications/base/models.py
```diff
@@ -444,9 +444,33 @@ def validate_global_setting(self):
             ):
                 raise ValidationError("There can only be one global setting per user.")
 
-    def save(self, *args, **kwargs):
-        if not self.web_notification:
+    def normalize_settings(self):
```
nemesifier, line 447: Also here I feel we need a docstring.

nemesifier (changes requested):
Great work @pandafy!
I did some manual testing and I confirm the 3 bugs are addressed by this PR! 👏 

Let's focus on polishing and stress testing the solution a bit further before merging.

## Clarify how tests map to bugs described in the issue

The issue mentions 3 bugs, I see 2 new tests in `openwisp_notifications/tests/test_notification_setting.py` and a bunch of tests in `openwisp_notifications/tests/test_notification_setting.py`, it's not obvious how those tests map to the bugs described in the issue, can you please clarify this?

It would help to explicitly flag the regression tests as such (maybe with a docstring).

## High level PR description missing

When we discussed this, we agreed to take a step back, describe the current model and design a new approach.

Can you please describe from a high level perspective the new model being introduced here?

## Other bugs found during testing (possibly unrelated)

Can you please confirm if you can replicate these? If they're not caused by these changes (seem like pre-existing issues to me), let's open other issues (1 for each).

- Deleting an `OrganizationUser(is_admin=True)` from a non-superuser does not remove the notification preferences of the user.

- Creating a staff user (but not superuser) with no related `OrganizationUser`, opening the notification preference page shows empty buttons,
  is this by design? Looks like a bug, would be better to at least allow changing the global preferences.

Image

----------

## Deep AI review

By Opus 4.7 and Kimi 2.5, please double check.

### `normalize_settings` is messy

`base/models.py:447-470` — three issues stacked in the same method:

- Typo: `should_enabled_web` should be `should_enable_web` (lines 452, 457, 469). Works but reads as a bug to anyone scanning the diff.
- The `if self.organization and hasattr(self.organization, "notification_settings")` at line 453 is redundant — we're already inside `if self.organization and self.type:` two lines above.
- The `if ... .web is not None` / `... .email is not None` checks (lines 456, 461) never short-circuit. `FallbackBooleanChoiceField.from_db_value` returns the fallback when the column is NULL, so the read attribute is never `None`. Same dead check is repeated in `email_notification` / `web_notification` properties (lines 516, 531). Just drop them.
- `if self.web_notification is False: self.email = self.web_notification` (lines 448-449) — assigning the property's value (which is always `False` here) is harder to read than `self.email = False`.

### `notify_handler` adds an extra query per send when there's a target_org

`handlers.py:74-81` does an `OrganizationNotificationSettings.objects.filter(...).only("web").first()` for every `notify.send` that has a target. For multi-recipient sends this isn't terrible (one query), but the org settings are also accessed lazily through `email_notification` later via `get_user_email_preference` — so the org row is loaded a second time per email. Either pass it through to the email path, or `select_related("organization__notification_settings")` on the recipient queryset and drop this manual lookup.

### `get_user_email_preference` has the same N+1 issue

`utils.py:170-173` does `notification.recipient.notificationsetting_set.get(...).email_notification`. The property then dereferences `self.organization.notification_settings`, which triggers a third query per notification when the post_save email handler fires. Add `select_related("organization__notification_settings")` to the `.get()` call.

### A global toggle overrides type-level defaults

Once the user touches the global setting, the bulk update writes explicit `True`/`False` onto every per-type row. That overrides the intent of types registered with `email_notification=False` (e.g. `generic_message`). Walkthrough: user starts with all rows at `email=None`. Property for `generic_message` resolves to False (type config). User toggles global email off, then on. After the second save, every row — including `generic_message` — has `email=True` stored explicitly. The property short-circuits on the explicit value and the user now gets emails for a type whose config says it shouldn't email by default. There's no UI hint that this happened, and the only way back is manually editing that specific row. This is pre-existing behavior, not introduced by this branch, but the PR's design intent ("harmonize on global change" per review-446.md) seems to be about org-level harmonization, not bulldozing type defaults. Worth deciding whether the propagation should set rows back to `None` (so type defaults reassert) instead of explicit booleans.

### Global save propagates `web` unconditionally

`base/models.py:480` always includes `"web": self.web` in `updates`, regardless of whether web actually changed. The companion email block (lines 489-490) is gated by `self.email != previous_state.email`. The asymmetry means: toggling **only** global email wipes out every per-type web override the user had set. If the design ("harmonize on global change") really wants full enforcement, both fields should be unconditional. If it wants to preserve overrides, both should be conditional. Pick one — the current shape is an accident. The change from `if not self.web` to `if self.web is False` is also a behavior change worth noting: a global with `web=None` (which can happen if someone clears it via the API) used to force email off, now it doesn't.

### Obsolete test assertions left behind

`tests/test_organization_setting.py::test_updating_notification_setting_for_deleted_org` still patches `NotificationSetting.objects.update` and asserts it wasn't called. Now that `_update_organizationuser_settings` is gone, that assertion is trivially true — no propagation path exists to call `.update()` from `OrganizationNotificationSettings.save()` anymore. Either remove the patch/assertion or repurpose the test.

### User's explicit "False" silently collapses to None

`normalize_settings` collapses both directions now: not just "matches type default" (old behavior) but "matches the resolved chain". So if org email=False and a user explicitly sets email=False on a specific type, it's stored as `None`. When the org later flips email back to True, that user starts receiving emails they had explicitly opted out of. The selenium test verifies this is intentional ("Return to inherited state → DB None"), but it does erase user intent. If you want user-explicit-False to survive an upstream flip, the collapse should only happen when the user value matches the *type default*, not the org-resolved value.

### `_set_org_notification_settings` test helper has a footgun

`tests/test_notifications.py:1556` — the helper signature defaults both `web` and `email` to `None`. Calling it as `_set_org_notification_settings(email=False)` would also write `web=None`, and `OrganizationNotificationSettings.save()` would then force `email=False` regardless because `if not self.web` treats None as falsy. No current test trips on this, but the helper invites bugs. Either make the args required or read current values and only overwrite what's passed.

### `create_notification_settings` logic is hard to follow

`tasks.py:90-109` — the email/web computation is correct but reads like a puzzle:

```python
email = False if global_setting.email is False else None
web = False if global_setting.web is False else None
if email is not False or web is not False:
    ...
    if email is not False:
        email = org_email
    if web is not False:
        web = org_web
if email is True:
    email = None
if web is True:
    web = None
```

The outer `or` gate, the per-field `is not False` re-checks inside, and the final True→None collapse all interact. Took a full trace to convince myself global-False is preserved. Either inline a comment explaining the three states (`False` = explicit deny, `None` = inherit, intermediate `True` = "matches default, collapse"), or split into a small helper that returns `(email, web)` from a clean if/elif chain.

### Selenium toast dismissal is racy

`tests/test_selenium.py::_dismiss_toast_if_present` waits up to 1s for `.toast`, then clicks it. If the toast is in the middle of fading out it'll be stale by the time `.click()` runs and you'll get a StaleElementReferenceException, not a TimeoutException. Wrap the click in a try/except for stale element, or click via JS to avoid the animation race.

### Minor: the comment about resolution order is slightly misleading

`handlers.py:91-97` says "Resolution order: user setting -> org setting -> type default". The actual filter is `web=True OR (web=None AND org_web is not False AND type_default is True)` — it's not really a chain, it's a combined condition. The comment is fine as a high-level summary but a fresh reader will look for chained code that isn't there.

### Pre-existing typo carried along

`base/models.py:486` — "email notifiations" in the comment. Not introduced by this PR but right next to the touched code, easy to fix while you're there.

pandafy:
> ## Other bugs found during testing (possibly unrelated)

I have opened respective issues 

> * Deleting an `OrganizationUser(is_admin=True)` from a non-superuser does not remove the notification preferences of the user.

https://github.com/openwisp/openwisp-notifications/issues/452

> * Creating a staff user (but not superuser) with no related `OrganizationUser`, opening the notification preference page shows empty buttons,
>   is this by design? Looks like a bug, would be better to at least allow changing the global preferences.

https://github.com/openwisp/openwisp-notifications/issues/453

docs/user/notification-preferences.rst
```diff
@@ -23,30 +23,55 @@ screenshot below:
...
 
     - You can disable notifications globally while still enabling them for
-      specific organizations.
+      specific organizations or notification types.
     - Notification settings are now linked: disabling web notifications
       will automatically disable email notifications, and enabling email
       notifications will automatically enable web notifications.
+    - Notification settings inherit organization and notification type
+      defaults whenever possible instead of storing redundant explicit
+      values.
```
nemesifier: I believe this is not something users care. Users want to use the preferences to reduce noise by reducing the amount notifications they receive.

openwisp_notifications/tests/test_notification_setting.py
```diff
@@ -493,6 +499,37 @@ def test_org_admin_addition_respects_global_preferences(self):
             self.assertEqual(org_setting.email, False)
             self.assertEqual(org_setting.web, None)
 
+    def test_org_setting_change_does_not_change_user_overrides(self):
+        """
+        Regression test for https://github.com/openwisp/openwisp-notifications/issues/448
```
nemesifier: There's 3 issues listed there, which bug does this address? Can you also clarify which other tests address the other bugs? We can refer to them as bug n.1/2/3.

nemesifier (changes requested):

nemesifier (approved):
Great work @pandafy!

Issue #452, opened 2026-05-11
[bug] Notification preferences not removed when removing OrganizationUser from organization admin


pandafy:
**Describe the bug**
Deleting an OrganizationUser (with is_admin=True) from a non-superuser does not remove the notification preferences for the user. As a result, former organization admins still have notification preferences associated with organizations they are no longer a member of.

**Steps To Reproduce**

1. Create an organization admin from the Django admin (ensure that this user is not organization owner).
2. Login as organization admin and go to http://127.0.0.1:8000/notifications/preferences/ to check notifications preferences. All notification preferences should be created as expected.
3. Delete the OrganizationUser (is_admin) object for the user created in step 1. This can be done from either Django shell or Django admin.
4. Login again as organization admin and go to http://127.0.0.1:8000/notifications/preferences/

**Expected behavior**
Notification preferences for this OrganizationUser should be deleted upon their removal from the organization.

**Actual outcome**
The preferences page still shows the preferences for the organization the user is no longer associated with.

Eeshu-Yadav:
picking up this one

BHARATH0153:
@Eeshu-Yadav are you working on it or can I proceed

Eeshu-Yadav:
> [@Eeshu-Yadav](https://github.com/Eeshu-Yadav) are you working on it or can I proceed

@BHARATH0153 it will be good as i am busy right now , can't pick up this one

BHARATH0153:
working on it

PR #456, opened 2026-05-12
[fix] Create global notification settings for non-org admin staff users #453

### LINKED ISSUE #453: [bug] Global notification preferences are not created for non-org admin staff users
### Bug Description
When a staff user (not a superuser) is created without any related `OrganizationUser`, accessing the notification preference page results in the page displaying empty buttons. 

#### Expected Behavior

The notification preference page should ideally allow such staff users to view and change at least global notification preferences, even if they are not associated with any `OrganizationUser`.

#### Actual Behavior

The notification preference page currently shows empty buttons with no actionable settings for staff users without an associated `OrganizationUser`.

Image

#### Steps to Reproduce
1. Create a staff user who is **not** a superuser and does **not** have a related `OrganizationUser`.
2. Log in as that user.
3. Open the notification preference page.
4. Observe that the buttons are empty and no preferences can be set.

prathmeshkulkarni-coder:
## Checklist

Closes #453.

## Description of Changes

Standard staff users (those who are not organization administrators) were
created without a global notification setting, leaving them with an empty
notification preferences UI. This PR ensures the global setting is created
and preserved for staff users across the relevant user and organization
lifecycle events.

### Handlers (`handlers.py`)

- The pre_save handler now reacts to changes in `is_staff` as well as
  `is_superuser`, and records transition flags on the user instance.
- The post_save handler reads those flags and dispatches the matching
  task on creation, privilege gain, or privilege loss, so staff users get
  a global setting on creation or promotion.

### Tasks (`tasks.py`)

- `create_superuser_notification_settings` now creates only the global row
  for staff users and is a no-op for regular users.
- `create_notification_settings` reactivates a soft-deleted global row when
  the user regains privilege, so re-promotions restore the UI.
- `superuser_demoted_notification_setting` and
  `update_org_user_notificationsetting` no longer wipe the global row when
  the user remains staff or still manages at least one organization. This
  also fixes a related bug where `exclude(organization_id__in=[...])` did
  not exclude rows where `organization_id IS NULL`, causing the global row
  to be soft-deleted when a staff user joined an organization.
- `ns_register_unregister_notification_type` ensures the global row exists
  for staff users on full registration. The staff loop is skipped on
  single-type registration since the global row is type-independent.

Public task and handler names were preserved for backward compatibility;
docstrings now note that they cover both staff users and superusers.

### Tests (`test_notification_setting.py`)

New regression tests cover:

- Staff user creation and promotion produce an active global setting.
- All demotion paths: staff to regular, superuser+staff to staff-only,
  and full privilege loss.
- Re-promotion reactivates a soft-deleted global setting.
- Joining an organization as a non-admin and as an admin both preserve a
  staff user's global setting.
- The pre_save transition flags do not leak across unrelated user saves.
- The task is a no-op when invoked for a regular user.

## Screenshot

N/A, backend-only change covered by tests.

### DIFF
--- a/openwisp_notifications/handlers.py
+++ b/openwisp_notifications/handlers.py
@@ -335,38 +335,52 @@ def notification_setting_delete_org_user(instance, **kwargs):
 @receiver(pre_save, sender=User, dispatch_uid="superuser_demoted_notification_setting")
 def superuser_status_changed_notification_setting(instance, update_fields, **kwargs):
     """
-    If user is demoted from superuser status, then
-    remove notification settings for non-managed organizations.
-
-    If user is promoted to superuser, then
-    create notification settings for all organizations.
+    Handles modification of user notification settings when
+    privileges change. This works on either staff users or super users.
     """
-    if update_fields is not None and "is_superuser" not in update_fields:
-        # No-op if is_superuser field is not being updated.
-        # If update_fields is None, it means any field could be updated.
+    if update_fields is not None and not {"is_superuser", "is_staff"}.intersection(
+        update_fields
+    ):
+        # No-op if relevant privilege fields are not being updated.
         return
     try:
-        db_instance = User.objects.only("is_superuser").get(pk=instance.pk)
+        db_instance = User.objects.only("is_superuser", "is_staff").get(pk=instance.pk)
     except User.DoesNotExist:
         # User is being created
         return
-    # If user is demoted from superuser to non-superuser
-    if db_instance.is_superuser and not instance.is_superuser:
-        transaction.on_commit(
-            lambda: tasks.superuser_demoted_notification_setting.delay(instance.pk)
-        )
-    elif not db_instance.is_superuser and instance.is_superuser:
-        transaction.on_commit(
-            lambda: tasks.create_superuser_notification_settings.delay(instance.pk)
-        )
+    was_privileged = db_instance.is_superuser or db_instance.is_staff
+    is_now_privileged = instance.is_superuser or instance.is_staff
+    # Transition Scenarios
+    gained_superuser = not db_instance.is_superuser and instance.is_superuser
+    gained_first_privilege = not was_privileged and is_now_privileged
+    lost_superuser = db_instance.is_superuser and not instance.is_superuser
+    lost_all_privileges = was_privileged and not is_now_privileged
+    # Attach transition flags to the instance to dispatch safely in post_save
+    # This bypasses eager Celery race conditions in testing environments
+    instance._lost_privileges = lost_superuser or lost_all_privileges
+    instance._gained_privileges = gained_superuser or gained_first_privilege
 
 
 @receiver(post_save, sender=User, dispatch_uid="create_superuser_notification_settings")
 def create_superuser_notification_settings(instance, created, **kwargs):
-    if created and instance.is_superuser:
+    """
+    Triggers initialization or modification of notification settings for privileged users.
+    Handles initial account creation as well as dynamic privilege level updates.
+    This works on either staff users or super users.
+    """
+    if created and (instance.is_superuser or instance.is_staff):
         transaction.on_commit(
             lambda: tasks.create_superuser_notification_settings.delay(instance.pk)
         )
+    elif not created:
+        if getattr(instance, "_lost_privileges", False):
+            transaction.on_commit(
+                lambda: tasks.superuser_demoted_notification_setting.delay(instance.pk)
+            )
+        elif getattr(instance, "_gained_privileges", False):
+            transaction.on_commit(
+                lambda: tasks.create_superuser_notification_settings.delay(instance.pk)
+            )
 
 
 @receiver(
--- a/openwisp_notifications/handlers.py
+++ b/openwisp_notifications/handlers.py
@@ -352,13 +352,11 @@ def superuser_status_changed_notification_setting(instance, update_fields, **kwa
         return
     was_privileged = db_instance.is_superuser or db_instance.is_staff
     is_now_privileged = instance.is_superuser or instance.is_staff
-    # Transition Scenarios
     gained_superuser = not db_instance.is_superuser and instance.is_superuser
     gained_first_privilege = not was_privileged and is_now_privileged
     lost_superuser = db_instance.is_superuser and not instance.is_superuser
     lost_all_privileges = was_privileged and not is_now_privileged
-    # Attach transition flags to the instance to dispatch safely in post_save
-    # This bypasses eager Celery race conditions in testing environments
+    # Flags are read by the post_save handler to dispatch the matching task.
     instance._lost_privileges = lost_superuser or lost_all_privileges
     instance._gained_privileges = gained_superuser or gained_first_privilege
 
--- a/openwisp_notifications/handlers.py
+++ b/openwisp_notifications/handlers.py
@@ -338,6 +338,8 @@ def superuser_status_changed_notification_setting(instance, update_fields, **kwa
     Handles modification of user notification settings when
     privileges change. This works on either staff users or super users.
     """
+    instance._lost_privileges = False
+    instance._gained_privileges = False
     if update_fields is not None and not {"is_superuser", "is_staff"}.intersection(
         update_fields
     ):
@@ -373,11 +375,15 @@ def create_superuser_notification_settings(instance, created, **kwargs):
             lambda: tasks.create_superuser_notification_settings.delay(instance.pk)
         )
     elif not created:
-        if getattr(instance, "_lost_privileges", False):
+        lost_privileges = getattr(instance, "_lost_privileges", False)
+        gained_privileges = getattr(instance, "_gained_privileges", False)
+        instance._lost_privileges = False
+        instance._gained_privileges = False
+        if lost_privileges:
             transaction.on_commit(
                 lambda: tasks.superuser_demoted_notification_setting.delay(instance.pk)
             )
-        elif getattr(instance, "_gained_privileges", False):
+        elif gained_privileges:
             transaction.on_commit(
                 lambda: tasks.create_superuser_notification_settings.delay(instance.pk)
             )
--- a/openwisp_notifications/tasks.py
+++ b/openwisp_notifications/tasks.py
@@ -156,7 +156,7 @@ def create_superuser_notification_settings(user_id):
             notification_types=types.NOTIFICATION_TYPES.keys(),
         )
     else:
-        # Creates global notification setting for the user
+        # Empty iterables create only the global setting row.
         create_notification_settings(user=user, organizations=[], notification_types=[])
 
 
--- a/openwisp_notifications/tasks.py
+++ b/openwisp_notifications/tasks.py
@@ -103,6 +103,9 @@ def create_notification_settings(user, organizations, notification_types):
     global_setting, _ = NotificationSetting.objects.get_or_create(
         user=user, organization=None, type=None, defaults={"email": True, "web": True}
     )
+    if global_setting.deleted:
+        global_setting.deleted = False
+        global_setting.save(update_fields=["deleted"])
 
     for type in notification_types:
         for org in organizations:
@@ -155,7 +158,7 @@ def create_superuser_notification_settings(user_id):
             organizations=Organization.objects.all(),
             notification_types=types.NOTIFICATION_TYPES.keys(),
         )
-    else:
+    elif user.is_staff:
         # Empty iterables create only the global setting row.
         create_notification_settings(user=user, organizations=[], notification_types=[])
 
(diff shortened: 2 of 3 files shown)


openwisp_notifications/handlers.py
```diff
@@ -332,40 +332,56 @@ def notification_setting_delete_org_user(instance, **kwargs):
...
-    # If user is demoted from superuser to non-superuser
-    if db_instance.is_superuser and not instance.is_superuser:
+
+    was_privileged = db_instance.is_superuser or db_instance.is_staff
+    is_now_privileged = instance.is_superuser or instance.is_staff
+
+    # Transition Scenarios
+    gained_superuser = not db_instance.is_superuser and instance.is_superuser
+    gained_first_privilege = not was_privileged and is_now_privileged
+
```
nemesifier: Too many blank lines, please avoid needless blank lines, use them sparingly because they do not provide much value.

openwisp_notifications/handlers.py
```diff
@@ -332,40 +332,56 @@ def notification_setting_delete_org_user(instance, **kwargs):
     )
 
 
-@receiver(pre_save, sender=User, dispatch_uid="superuser_demoted_notification_setting")
-def superuser_status_changed_notification_setting(instance, update_fields, **kwargs):
+@receiver(
+    pre_save, sender=User, dispatch_uid="user_status_changed_notification_setting"
+)
+def user_status_changed_notification_setting(instance, update_fields, **kwargs):
```
nemesifier: Unfortunately we can't rename this if we want to maintain backward compatibility, please avoid the renaming and add a brief docstring point out this works on either staff users or super users.

openwisp_notifications/handlers.py
```diff
@@ -332,40 +332,56 @@ def notification_setting_delete_org_user(instance, **kwargs):
...
+                instance.pk, is_superuser=instance.is_superuser
+            )
         )
 
 
-@receiver(post_save, sender=User, dispatch_uid="create_superuser_notification_settings")
-def create_superuser_notification_settings(instance, created, **kwargs):
-    if created and instance.is_superuser:
+@receiver(post_save, sender=User, dispatch_uid="create_user_notification_settings")
+def create_user_notification_settings(instance, created, **kwargs):
```
nemesifier: Unfortunately we can't rename this if we want to maintain backward compatibility, please avoid the renaming and add a brief docstring point out this works on either staff users or super users.

openwisp_notifications/tasks.py
```diff
@@ -142,29 +142,44 @@ def create_notification_settings(user, organizations, notification_types):
 
 
 @shared_task(base=OpenwispCeleryTask)
-def create_superuser_notification_settings(user_id):
+def create_user_notification_settings(user_id, is_superuser=None):
```
nemesifier: Unfortunately we can't rename tasks if we want to maintain backward compatibility, please avoid the renaming and add a brief docstring point out this works on either staff users or super users.

openwisp_notifications/tasks.py
```diff
@@ -142,29 +142,44 @@ def create_notification_settings(user, organizations, notification_types):
...
-    NotificationSetting.objects.filter(user_id=user_id).exclude(
+    qs = NotificationSetting.objects.filter(user_id=user_id).exclude(
         organization__in=user.organizations_managed
-    ).update(deleted=True)
+    )
+
+    # Do not delete global setting if user is staff or manages at least one organization
+    if user.is_staff or user.organizations_managed:
+        qs = qs.exclude(organization=None)
+
```
nemesifier: Remove the blank lines

nemesifier (changes requested):
Great work @prathmeshkulkarni-coder, the solution works, now we need to polish it a little bit.

## Critical bug, the new global setting gets wiped when the staff user joins any org

This is the most important finding and it must be fixed before merging.

`openwisp_notifications/tasks.py:230-236` in `update_org_user_notificationsetting`
runs whenever an `OrganizationUser` is saved (signal in
`openwisp_notifications/handlers.py:313-321`):

```python
if not user.is_superuser:
    NotificationSetting.objects.filter(user=user).exclude(
        organization_id__in=user.organizations_managed
    ).update(deleted=True)
```

The `exclude(organization_id__in=...)` does not preserve the global row
(`organization_id IS NULL`):

- When `organizations_managed` is empty (staff user not admin anywhere),
  Django optimises `exclude(field__in=[])` into a query with no filter, so
  the query matches every row for the user, including the global one, and
  marks them all deleted.
- When `organizations_managed` is non-empty, Django emits
  `NOT (org_id IN (...) AND org_id IS NOT NULL)`. For the row with
  `org_id = NULL`, the inner AND short-circuits to FALSE, the outer NOT
  becomes TRUE, and the global row matches the WHERE clause and gets
  `deleted=True`.

I reproduced both with the test settings (CELERY_TASK_ALWAYS_EAGER=True):

1. Create a staff user → global row exists, `deleted=False`. ✓
2. Add that user to any organization (admin or not) → global row is now
   `deleted=True`. ✗

This means the fix only works for staff users who are not associated with
any organization at all. The moment they are added to an org, which is a
normal operation, the empty-buttons bug is back. No test in this PR covers
this flow, which is why it slipped through.

The fix has to happen inside `update_org_user_notificationsetting`. The same
pattern used in the demotion task, explicitly excluding `organization=None`
when the user should retain the global row, works here too. Something like:

```python
qs = NotificationSetting.objects.filter(user=user).exclude(
    organization_id__in=user.organizations_managed
)
# never delete the global setting for staff or for users still managing orgs
if user.is_staff or user.organizations_managed:
    qs = qs.exclude(organization=None)
qs.update(deleted=True)
```

Please also add a regression test that creates a staff user, attaches them
to an org both as `is_admin=True` and `is_admin=False`, and asserts the
global row stays `deleted=False` in both cases.

## CodeRabbit's two flagged scenarios, already covered, but no tests

CodeRabbit flagged two demotion paths as broken. After tracing the new logic
they are actually handled correctly:

- superuser+staff → staff: hits `lost_superuser`, calls the demote task,
  which keeps the global because `user.is_staff` is True. Confirmed by
  running it locally, global stays alive, per-org rows go away.
- staff → regular: hits `lost_all_privileges`, calls the demote task,
  global gets removed because `user.is_staff` is False and the user manages
  nothing. That matches the intent.

So CodeRabbit's claim is wrong about the behaviour, but it is right that
neither path has a test. Please add tests for both, otherwise this is
fragile to future refactors.

## Smaller issues

### The `is_superuser` kwarg passed into the task is redundant

`create_user_notification_settings(user_id, is_superuser=None)` falls back
to `user.is_superuser` from the DB when `is_superuser` is not passed. But
the call sites always pass it, and the DB value at task time is identical
to what gets passed (the task runs under `transaction.on_commit`, after the
save has committed). The PR description says this exists to "statically
bypass edge-case race conditions in non-deterministic asynchronous queuing
pipelines", but there is no such race here: the task already loads the user
record by primary key. Drop the parameter and the fallback branch and let
the task read `user.is_superuser` directly, less surface area, same
correctness.

### Staff loop in `ns_register_unregister_notification_type` runs for every type registration

`openwisp_notifications/tasks.py:203-205` adds:

```python
for user in User.objects.filter(is_staff=True, is_superuser=False).iterator():
    create_notification_settings(user, [], [])
```

`create_notification_settings(user, [], [])` does nothing except a
`get_or_create` for the global row. That's fine on full registration. But
this task is also called once per type when types are registered one by one
(`ns_register_unregister_notification_type(notification_type=X)`), and it
runs the same N-staff-users loop each time even though there is nothing to
do after the first run. It is harmless but wasteful, consider skipping
the staff loop when `notification_type` is not None, since the global row
is independent of types and only needs to be ensured once.

### PR description has clearly LLM-generated phrasing

Lines like "statically bypass edge-case race conditions in non-deterministic
asynchronous queuing pipelines" and "Regressional Coverage: Introduced
dedicated verification suites" don't describe anything real. The
description should plainly say what was changed and why, same staff-user
creation flow, same handler additions, two new tests. Worth tidying up
before merge so the git log is readable.

### Existing test assertion is loosened without justification

`test_task_not_called_on_user_login` changes from
`created_mock.assert_called_once()` to
`self.assertEqual(created_mock.call_count, 2)`. That number happens to be
correct, admin counts once, staff org admin counts once, but the test
no longer expresses *why* it expects 2. A short comment, or splitting the
assertion per user, would make it less brittle.

### Other observations

See my inline comments below.

nemesifier:
Pushed one small follow-up fix.

While doing one last pass I noticed the privilege transition flags were stored on the User instance and could leak if the same instance was saved again later with unrelated update_fields. That could enqueue the create/demote notification setting task again when nothing privilege-related changed.

I reset those flags before computing transitions and clear them after post_save consumes them. Added a regression test covering both gained and lost privilege transitions followed by an unrelated save. Also clarified the global-row comment in the task.

All good now.

nemesifier:
@coderabbitai resolve

nemesifier:
@coderabbitai full review

nemesifier:
@prathmeshkulkarni-coder I did one more pass on the tests.

I temporarily reverted the implementation changes while keeping the new/changed tests, then ran only those tests to check whether each one actually catches the patch. All of them failed without the fix except `test_superuser_and_staff_demoted_to_regular_user`, which was still passing on the old implementation.

I removed that test for now because, as written, it does not protect this change. Was that intentional? Were you aware it was already passing without the fix? Are you sure we need that coverage here, or was the scenario useful but the test not written in a way that exercises the actual regression?

I also removed the extra blank lines / boilerplate docstrings in the added tests to keep this file more compact.

prathmeshkulkarni-coder:
@nemesifier , I added test_superuser_and_staff_demoted_to_regular_user entirely because the Coderabbit specifically flagged that exact transition as missing explicit test coverage in its review. I was aware the logic handled it correctly, but I just added it to satisfy the bot's coverage demands. I am totally fine with you removing it to keep the file compact.
And sorry about blank lines i was just confuse which one to remove now i understand :)

nemesifier:
@prathmeshkulkarni-coder I tried `test_superuser_and_staff_demoted_to_regular_user` manually and it seems buggy to me: notification settings are not removed for the user.

prathmeshkulkarni-coder:
@nemesifier , If the test is buggy locally, then I completely agree we should remove it. As you mention yesterday, if a `test_superuser_and_staff_demoted_to_regular_user ` isn't accurately protecting a regression and just adds noise, it's better to leave it out. 

**We already have tests for a superuser being demoted to staff (`test_superuser_and_staff_demoted_to_staff_only`), and tests for a staff member being demoted to a regular user (`test_staff_demoted_to_regular_user`). The core logic is already proven to work.**

Should I try to rewrite the test in a different way to properly capture that transition, or are we good to proceed without it?

nemesifier:
@prathmeshkulkarni-coder I did not mean to say the test was buggy. I meant to say the bug covered by the test happens, I can replicate it with manual testing by demoting a user from staff user to normal user and the notification preferences do not get deleted.

PS: I am investigating, I think it's due to #460 missing here, now that I merged with master it's better.

prathmeshkulkarni-coder:
@nemesifier , i just reproduced it locally.

The issue was in tasks.py. When a staff user who manages no organizations is demoted, user.organizations_managed evaluates to an empty list []. Passing an empty list to exclude(organization__in=[]) on a nullable field causes the Django ORM to silently skip updating the global setting on some database backends (this is actually the exact ORM quirk that CodeRabbit was trying to warn me about earlier!).

I am working on a commit that wraps the exclude(organization__in=...) queries in an explicit if user.organizations_managed: check in both tasks. I also added the test back in since it is clearly very valuable for catching this specific regression

nemesifier:
> @nemesifier , i just reproduced it locally.
> 
> The issue was in tasks.py. When a staff user who manages no organizations is demoted, user.organizations_managed evaluates to an empty list []. Passing an empty list to exclude(organization__in=[]) on a nullable field causes the Django ORM to silently skip updating the global setting on some database backends (this is actually the exact ORM quirk that CodeRabbit was trying to warn me about earlier!).
> 
> I am working on a commit that wraps the exclude(organization__in=...) queries in an explicit if user.organizations_managed: check in both tasks. I also added the test back in since it is clearly very valuable for catching this specific regression

@prathmeshkulkarni-coder doesn't seem to be the case, it was probably due to #452 that I was still seeing the notification preferences for a demoted user.

nemesifier (approved):
Should be ready, flagging as approved, doing some last checks in the meantime.

Opened https://github.com/openwisp/openwisp-notifications/issues/462 for the other issue I found.

prathmeshkulkarni-coder:
> @prathmeshkulkarni-coder doesn't seem to be the case, it was probably due to https://github.com/openwisp/openwisp-notifications/issues/452 that I was still seeing the notification preferences for a demoted user.

Understood! Good to know it was just an issue with #452. I've discarded my local commit, so the code is good to go whenever you are ready to merge.

PR #463, opened 2026-05-15
[fix] Restrict notification preferences page to admins #462

### LINKED ISSUE #462: [bug] Restrict notification preferences page for regular users
**Describe the bug**

When logged as a staff/superuser, the notification preferences page for non-staff, non-superuser can be opened directly if the URL is known, even though the "notification preferences" button correctly disappears from that user's admin change page.

The page should not be accessible if the user being edited is not staff or superuser.
In this state the page looks broken and exposes UI which should not be available for end-users.

This was noticed while reviewing #456, but it looks like a separate access-control/UI issue and should be fixed independently.

Please start by adding a failing regression test first, then implement the fix.

**Steps To Reproduce**

1. Create or use a regular user with `is_staff=False` and `is_superuser=False`.
2. Open the Django admin change page for that user while being logged as staff or superuser (both cases must be covered in 2 separate tests)
3. Confirm the notification preferences button is not shown.
4. Open that user's notification preferences URL directly.
5. The page still opens and looks broken.

**Expected behavior**

The notification preferences page should only open when the user being edited is staff or superuser.

If the target user is neither staff nor superuser, the view should redirect to the admin index and show an error message using Django's messages framework if possible.

**Screenshots**

Not available.

**System Informatioon:**

- OS: Ubuntu/Linux development environment
- Python Version: Not specific to a Python version
- Django Version: Not specific to a Django version
- Browser and Browser Version (if applicable): Not browser-specific

nemesifier:
## Checklist

Closes #462.

## Description of Changes

Restricts direct access to the user notification preferences page so it only opens for staff users and superusers.

If the target user is a regular non-staff account, the view redirects to the admin index and shows a translatable error message explaining that notification preferences are only available for staff users or superusers.

Added regression coverage for the redirect/message behavior and updated the existing preference page test to use a staff target user for the allowed access case.

## Screenshot

Not applicable: this is a redirect/error-message behavior covered by tests.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -31,6 +31,7 @@ The repository is a single Python package, not a monorepo.
 - Use `.github/workflows/build.yml` as the authoritative compatibility matrix for supported Python and Django versions.
 - Preserve swappable model support and integration with `openwisp-users` organizations and memberships.
 - Be careful with notification preference inheritance, global notification settings, soft-deleted `NotificationSetting` rows, and user/organization permission boundaries.
+- Mark user-facing strings as translatable with Django i18n helpers, typically `gettext_lazy` imported as `_`.
 - Add or update tests for behavior changes.
 - For bug fixes, prefer the red/green workflow: write or update the regression test first, run it against the unfixed code and confirm the failure message is clear, then apply the fix and rerun the targeted test plus the relevant module.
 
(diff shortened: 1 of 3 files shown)


nemesifier:
@coderabbitai a selenium browser test here is not extremely needed because the redirect is implemented in the server side which is perfectly testable with python unit tests. We need selenium tests for JS centric functionality.

nemesifier:
@coderabbitai any outstanding issues to resolve? If not flag as approved.

PR #479, opened 2026-06-04
[fix] Fixed accidental email delivery for generic_message notifications #478

### LINKED ISSUE #478: [bug] Unexpected email delivery for generic_message notifications on shared objects
**Describe the bug**

`get_user_email_preference()` in `openwisp_notifications/utils.py` returns `True` when the notification's target has no organization (`organization_id` is `None`, typical for shared objects). This causes email notifications to be sent even when the notification type's `email_notification` configuration is `False`, e.g. for `generic_message`.

**Steps To Reproduce**

1. Create a notification with a valid type (e.g. `generic_message`, which has `email_notification: False`) and a target whose `organization_id` is `None`.
2. Observe that `get_user_email_preference()` returns `True` because the guard clause `if not (notification.type and target_org): return True` treats a missing `target_org` the same as a missing `type`.
3. An email notification is sent despite the type default being `email_notification: False`.

**Expected behavior**

When `organization_id` is `None` (shared object) but the notification has a valid `type`, the function should check the notification type's `email_notification` configuration instead of blindly returning `True`. For `generic_message` (which defaults to `email_notification: False`), no email should be sent.

pandafy:
## Checklist

- N/A I have updated the documentation.

Fixes #478

### DIFF
--- a/openwisp_notifications/utils.py
+++ b/openwisp_notifications/utils.py
@@ -164,6 +164,8 @@ def get_user_email_preference(notification):
     """
     target_org = getattr(getattr(notification, "target", None), "organization_id", None)
     if not notification.type:
+        # notify.send() may create notifications without a type; when no type is
+        # available, default to email notifications being enabled.
         return True
     if not target_org:
         type_config = get_notification_configuration(notification.type)
--- a/openwisp_notifications/utils.py
+++ b/openwisp_notifications/utils.py
@@ -14,6 +14,7 @@ from django.utils.translation import ngettext_lazy
 from openwisp_notifications import settings as app_settings
 from openwisp_notifications.exceptions import NotificationRenderException
 from openwisp_notifications.swapper import load_model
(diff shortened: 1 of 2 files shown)


openwisp_notifications/utils.py
```diff
@@ -158,14 +159,15 @@ def send_notification_email(
...
-    If the user has no preference set, it defaults to True.
+    Falls back to the notification type's ``email_notification``
+    configuration when the target has no organization.
     """
     target_org = getattr(getattr(notification, "target", None), "organization_id", None)
-    if not (notification.type and target_org):
-        # We can not check email preference if notification type is absent,
-        # or if target_org is not present
-        # therefore send email anyway.
+    if not notification.type:
```
nemesifier, line 166: When can this happen? A comment would be helpful here.

nemesifier (approved):

PR #493, opened 2026-07-18
[feature] Added target_url_suffix param to notify signal #492

### LINKED ISSUE #492: [feature] Add `target_url_suffix` for notification target URLs
**Is your feature request related to a problem? Please describe.**

We need a way to append a querystring or fragment to a notification target URL, especially for `generic_message` notifications, without allowing callers to override the full URL.

**Describe the solution you'd like**

Add a `target_url_suffix` argument to `notify.send`, for example:

```python
notify.send(
    type="generic_message",
    sender=obj,
    target=obj,
    target_url_suffix="?status=pending",
)
```

The suffix should be appended in the notification model when resolving `target_url`. It should also support fragments, for example `target_url_suffix="#pending"`.

Invalid values should raise `ValueError` when passed to `notify.send`. Valid suffixes should start with `?`, `&`, or `#`.

**Describe alternatives you've considered**

Using the existing notification type `target_link` mechanism works for dedicated notification types, but it is type-level and not suitable for per-notification `generic_message` usage.

Allowing a full URL override is too broad for this use case.

**Additional context**

The goal is to keep the feature narrow and model-centered: generate the existing target URL as usual, then append the validated suffix.

Originally needed in https://github.com/openwisp/openwisp-firmware-upgrader/pull/436.

nemesifier:
## Checklist

Closes #492

## Description of Changes

This PR adds a way to append a querystring or fragment to a notification target URL, especially for `generic_message` notifications, without allowing callers to override the full URL. It does so by adding support for a new argument `target_url_suffix` argument to `notify.send`.

### DIFF
--- a/docs/developer/notification-types.rst
+++ b/docs/developer/notification-types.rst
@@ -57,6 +57,24 @@ type:
     software like Aldus PageMaker including versions of *Lorem Ipsum*.""",
     )
 
+The URL of a ``generic_message`` notification points to the target object
+by default. To append a querystring or fragment to this URL, pass
+``target_url_suffix`` to ``notify.send``:
+
+.. code-block:: python
+
+    notify.send(
+        type="generic_message",
+        level="error",
+        message="An unexpected error happened!",
+        sender=User.objects.first(),
+        target=User.objects.last(),
+        target_url_suffix="?status=pending",
+    )
+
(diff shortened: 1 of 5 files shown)


nemesifier:
@coderabbitai resolve

